PR #509, opened 2015-06-16
Optimize

boxerab:
I have added some timing code to opj compress and decompress., plus a few small T1 optimizations.
Also, a windows fix - visual studio lrintf is extremely slow, doubling decode time. Keep using old
code for lrintf.


boxerab:
Compiler should take care of some of these optimizations: need to re-submit request.

PR #511, opened 2015-06-16
Optimize

boxerab:
Here are a few small T1 optimizations: removing some unnecessary calculations.
Also, there is a patch for lrintf on windows - the visual studio version has problems 
and doubles the decode time. I have also added some timing code to compress and decompress.

### DIFF
--- a/src/bin/jp2/opj_compress.c
+++ b/src/bin/jp2/opj_compress.c
@@ -1547,7 +1547,7 @@ OPJ_FLOAT64 opj_clock(void) {
 	/* cout << "freq = " << ((double) freq.QuadPart) << endl; */
     /* t is the high resolution performance counter (see MSDN) */
     QueryPerformanceCounter ( & t ) ;
-    return ( t.QuadPart /(OPJ_FLOAT64) freq.QuadPart ) ;
+    return freq.QuadPart ? ( t.QuadPart /(OPJ_FLOAT64) freq.QuadPart ) : 0 ;
 #else
 	/* Unix or Linux: use resource usage */
     struct rusage t;
@@ -1871,8 +1871,9 @@ int main(int argc, char **argv) {
     if(raw_cp.rawComps) free(raw_cp.rawComps);
 	
 	t = opj_clock() - t;
-	fprintf(stdout, "encode time: %d ms \n", (int)((t * 1000)/num_compressed_files));
-	scanf("%d");
+	if (num_compressed_files)
+		fprintf(stdout, "encode time: %d ms \n", (int)((t * 1000)/num_compressed_files));
+	//getch());
 
     return 0;
 }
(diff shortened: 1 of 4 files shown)


detonin:
Hi Aaron, many thanks for these optimizations. Did you already run the test suite on the modified code ? If you enable testing in cmake options, download the test data and run "make test", you will have all tests processed. Some of them are failing but could you check that it does not induce regression compared to the situation here: http://my.cdash.org/index.php?project=OPENJPEG. This would certainly speed up merging of your contributions. Contact me if you have any issue with processing the tests.

And if the same could be done for the other PR, that would be great. Thanks !

boxerab:
Hi Antonin,
Thanks for your email.  Can I run the test suite from windows?  I tried running the RUN_TESTS project, but it did nothing.

Cheers,
Aaron

mayeut:
Hi @boxerab,
You can run the test suite from windows.
Here are the cmake options I use when working under Visual Studio:
BUILD_TESTING=TRUE
OPJ_DATA_ROOT=C:\test\openjpeg-data (uclouvain/openjpeg-data master)
BUILD_THIRDPARTY=TRUE

You need to build ALL_BUILD before building RUN_TESTS.

boxerab:
Thanks, Matthieu. Unfortunately, this does not work.  I get the message

Unable to start program
'C:\src\openjpeg-build\x64\Release\RUN_TESTS'
Access is denied

mayeut:
OK, that's weird never ran into that one.

Did you try to Run/Debug (F5) or just build ?
You need to only build the project. Running/Debugging won't work.

boxerab:
OK, so if I build the project, then the tests run? Where do I see the results?

boxerab:
Yeah, the error occurred when I tried to run the project.

boxerab:
ahhhh, ok, I got it. Working now.

boxerab:
I get some "TIfError cannot open" error dialogs. Is this normal?

boxerab:
never mind - look like I need to fix some things.

boxerab:
Tests are awesome - I found and fixed a few bugs on this pull request. Now failing same number of tests as master - 26.

src/bin/jp2/opj_compress.c
```diff
@@ -1535,6 +1538,31 @@ static void info_callback(const char *msg, void *client_data) {
     fprintf(stdout, "[INFO] %s", msg);
 }
 
+OPJ_FLOAT64 opj_clock(void) {
```
detonin: Is there a reason why you did not use opj_clock.c and opj_clock.h already defined in the repo ? If possible, I would prefer these functions to be used rather than rewriting them in opj_(de)compress. Could you update the PR in that way ?

boxerab: I would be happy to re-use the code, but opj_clock is not exported by the library, so sadly I don't have access to it from opj_(de)compress..

boxerab:
Closing - these changes are all included in T1 Optimize pull request. (#513)

PR #513, opened 2015-06-19
T1 optimize

boxerab:
I have integrated Carl's T1 patch into my OpenMP branch.  This gives a roughly 22% increase in encoding speed.


boxerab:
Test results now equivalent to master (26 failed)

tverbeke:
+1

boxerab:
Actually, with OpenMP turned on, I am only seeing a 10% reduction in encoding time using this patch, relative to OpenMP without patch.

boxerab:
OK, I am happy with this patch - not planning on making any changes to this branch.

boxerab:
Note: the usage in this pull request of SSE on x64 to fix the windows lrintf bug is justified by the fact that all x64 intel architectures support SSE and SSE2.

http://stackoverflow.com/questions/28185844/do-all-64-bit-intel-architectures-support-ssse3-sse4-1-sse4-2-instructions

detonin:
Hi Aaron, many thanks for these optimizations. For now, I cherry-picked the first four commits. I'd like some more testing before including OpenMP. A student here at our lab is working on multi-threading openjpeg at code-block level ... Would this make OpenMP useless ?

Concerning Carl's T1 optmization, I'm currently not in favor of adding new t1_opt files. I think we should keep things simple and have a single t1 implementation. What do you think ? 

Concerning cherry-picking pull requests, I don't know if it's the best way to go, I'm still in the github learning curve. Let me know if you have advices on this. In any case, could you update your pull request or let me know hox do I do that ? 

Thanks

boxerab:
Hi Antonin,

Cherry picking is fine; you should still be able to get the entire pull request at a later date, without any problems.

Regarding your student: yes: OpenMP does multi-threaded openjpeg at code block level. And OpenMP is platform independant, so I think it is a very nice solution. Not much code has to change.
Also, the OpenMP branch multi-threads  DWT across components. And, it supports mutli-threading across images, for batch encoding/decoding, so it allows users to use 100% of their cores.

Perhaps the student could look at DWT boundary extension instead ? :)

As for the t1_opt file, certainly we could put everything into one file. Personally, I like a more modular approach - easier to work with the code. The library already has at least one +10,000 line file.
This is usually a sign that refactoring is needed.  In my experience.

Cheers,
Aaron

Issue #254, opened 2014-01-29
Loss decoding quality in 2.0.0


gcode-importer:
Originally reported on Google Code with ID 254
```
> What steps will reproduce the problem?

Try to convert the attached image with opj_decompress (fresh sources) into a png one.

> What is the expected output? 

Converted image with good quality, like generated by j2k_to_image from 1.5.1.

> What do you see instead?

Bad image quality instead

> What version of the product are you using? On what operating system?

actual SVN sources on Mint Maya (Linux 3.2.0-40-generic-pae #64-Ubuntu  i686 i686 i386
GNU/Linux)

> Please provide any additional information below.

The attached image is extracted from a PDF file which normally shown by MuPDF engine
before its migration from 1.5.1 to 2.0.0

Possibly this issue has the same cause with #252, but no error messages written on
decoding.


```

Reported by `Alexander.V.Kasatkin` on 2014-01-29 15:23:09

<hr>
 * *Attachment: [1.jp2](https://storage.googleapis.com/google-code-attachments/openjpeg/issue-254/comment-0/1.jp2)*
 * *Attachment: 2.0.0.png<br>![2.0.0.png](https://storage.googleapis.com/google-code-attachments/openjpeg/issue-254/comment-0/2.0.0.png)*
 * *Attachment: 1.5.1.png<br>![1.5.1.png](https://storage.googleapis.com/google-code-attachments/openjpeg/issue-254/comment-0/1.5.1.png)*

PR #514, opened 2015-06-19
Correctly decode files with incorrect tile-part header fields (TPsot==TNsot)

mayeut:
fixes uclouvain/openjpeg#254


mayeut:
@detonin ,

All tests OK (see #254 for details). Given all comments in #254, I think it's OK to merge.
I can add a build option (Not to change API/ABI at all) do disable/enable the check.
I can add a new function to enable/disable the check.
If any of the above are needed, we shall choose what's the default value

szukw000:
This patch for issue254 has a drawback. I downloaded openjpeg-master on 2015-06-23 and found
that my viewer showed a correct JP2 image for the respective image.

But the decompressed images (PNG, BMP, etc) are all unsharp and washed out as the 'original' JP2
image.

The last patch from mdarbois was OK: JP2 and decompressed images are all sharp.

winfried

detonin:
@mayeut Thanks for the PR. For now, I propose you add a build option which would be enabled by default. Once we change API/ABI, I suggest to add a commandline option, such as "--robust" or sth like that to enable robustness against corrupted codestreams (at a price of small performance decrease). By the way, this PR triggers 4 additional MD5 failing tests. Corresponding MD5 files shall be updated accordingly ... Could you also take care of that ? Many thanks;

@szukw000 Could you give a link to the corresponding JP2 image ?

szukw000:
The file:

data/input/nonregression/issue254.jp2

Broken opj_decompress in:

openjpeg-git-2015-06-29

Perfect opj_decompress in:

openjpeg-2.x-trunk-r3007 patched with:

For those interested,
Updated patch for r3007

Reported by mayeut on 2015-06-05 19:03:36

issue254-r3007.patch

winfried

szukw000:
In openjpeg-git-2015-06-29 I have now included 'flviewer'. And found that the
file 'issue254.jp2' is broken as before: unsharp and washed out.

That my viewer shows a sharp JP2 image does have a simple reason: the viewer 
uses the  'issue254-r3007.patch'.

Does the above mentioned fix  really fix the cyrillic image of issue254?

winfried

mayeut:
@szukw000 ,

The issue254-r3007.patch is this PR. It was created by a git diff between master (synced to r3007 thanks to @detonin ) & my issue-254 branch.
I'll make a final check with the up to date master before committing but I don't see why it wouldn't work as expected.
Here's the image I get so far (which is sharp & the correct one I think)
issue254-fix

mayeut:
@detonin 
1 test less failing after this PR (MD5 for issue254 was wrong previously).
That's 26 tests failing on my setup. When disabling the fix, goes up to 31 (4 already mentioned in uclouvain/openjpeg#254 + issue254 MD5).

szukw000:
@mayeut ,
yes: the patch for r3007 is correct and perfect. I misunderstood the first comment
on this page and thought that there was another, later patch.

winfried

rouault:
Behavior of this PR will be partly changed by https://github.com/uclouvain/openjpeg/pull/1560

PR #519, opened 2015-06-25
libtiff Version 4.0.4beta

boxerab:
We have a three year old version of libtiff.  This commit upgrades libtiff to version 4.04 beta.  
As you can see in the code, there are quite a few potential memory leaks plugged and other bug fixes.


detonin:
Thanks. I upgraded to recently released 4.0.4 instead of 4.0.4beta.

Issue #518, opened 2015-06-24
Bigendian: opj_compress + opj_decompress fails


manisandro:
On bigendian, converting some image to j2k and decompressing it again results in failure, i.e.:

    $ wget http://people.sc.fsu.edu/~jburkardt/data/ppma/star_field.ascii.ppm
    
    $ opj2_compress -i star_field.ascii.ppm -o test.j2k -q 30,40,50
    [INFO] tile number 1 / 1
    Generated outfile test.j2k
    
    $ opj2_decompress -i test.j2k -o test.ppm
    [INFO] Start to read j2k main header (0).
    [ERROR] Prevent buffer overflow (x1: 671090818, y1: 671090818)
    [ERROR] Marker handler function failed to read the marker segment
    ERROR -> opj_decompress: failed to read the header

A quick scan through the codebase reveals for instance in `convert.c` portions of code such as

                union { signed short val; signed char vals[2]; } uc16;
                mask = (1 << image->comps[compno].prec) - 1;
                ptr = image->comps[compno].data;
                for (line = 0; line < h; line++) {
                    for(row = 0; row < w; row++)	{
                        curr = *ptr;
                        if(curr > 32767 ) curr = 32767; else if( curr < -32768) curr = -32768;
                        uc16.val = (signed short)(curr & mask);
                        res = fwrite(uc16.vals, 1, 2, rawFile);
                        if( res < 2 ) {
                            fprintf(stderr, "failed to write 2 byte for %s\n", outfile);
                            goto fin;
                        }
                        ptr++;
                    }
                }

which likely produce wrong results on bigendian.

sharkcz:
Let me know you will need access to a big endian machine (ppc64)

szukw000:
On Wed, 24 Jun 2015 16:44:50 -0700, Sandro Mani wrote:

>On bigendian, converting some image to j2k and decompressing 
>it again results in failure, i.e.:

A long time ago I sent a patch for PPC64 to openjpeg@googlegroups.com
for this BIGENDIAN bug. This patch has been ignored.

Here it is again. Ignore it once more. Or apply it.

winfried

--------------------------------------------------

--- openjpeg-2.x-trunk-r3006-1/src/lib/openjp2/cio.c.orig	2015-05-31 23:37:08.000000000 +0000
+++ openjpeg-2.x-trunk-r3006-1/src/lib/openjp2/cio.c	2015-06-01 01:54:50.503990204 +0000
[...]

PR #521, opened 2015-06-25
Fix opj_write_bytes_BE (#518)

manisandro:
This patch, originally by @szukw000, fixes the issue where opj_compress + opj_decompress fails on bigendian systems.

### DIFF
--- a/src/lib/openjp2/cio.c
+++ b/src/lib/openjp2/cio.c
@@ -46,7 +46,7 @@
 
 void opj_write_bytes_BE (OPJ_BYTE * p_buffer, OPJ_UINT32 p_value, OPJ_UINT32 p_nb_bytes)
 {
-	const OPJ_BYTE * l_data_ptr = ((const OPJ_BYTE *) &p_value) + p_nb_bytes;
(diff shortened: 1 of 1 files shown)


detonin:
@manisandro Many thanks for the PR.
@sharkcz I do not have access to any big endian machine: is there a way you could run the test suite with and without the patch and report results ? To run the test suite, you have to clone openjpeg-data repository, enable testing in cmake configuration, and indicate to cmake where to find openjpeg-data folder. Then, "make & make Experimental" launches the test suite and submits results to CDash dashboard (http://my.cdash.org/index.php?project=OPENJPEG). Let me know.

mayeut:
This is related to #518 (already in the title but needed to properly create a link between the 2 it seems)

szukw000:
@detonin, I have done both.

Site name: G5
Build name: Linux-(g5-3.16.0-4-powerpc64-#1-SMP-Debian-3.16.7-ckt11-1-(2015-05-24)-ppc64-GNU/Linux)

The first is without, the second with, the cio.c patch of #518.

winfried

PR #528, opened 2015-07-04
Update zlib to version 1.2.8

mayeut:
This updates lib to 1.2.8
Tested on MacOS X OK (I tested with CMakeLists.txt in bin/jp2 modified not to link against System library - default on APPLE even when building with third party)
If someone can test this PR on other platforms, would be great


mayeut:
Tested also on 
Win7 x64 Visual 2015
Ubuntu 14.04 x86_64
Ubuntu 14.04 ppc64

Issue #447, opened 2014-12-04
Image in pdf don't display


gcode-importer:
Originally reported on Google Code with ID 447
```
What steps will reproduce the problem?
1. Open the attached document in chrome
2. Image on page 11 is black

This issue is from 
https://code.google.com/p/chromium/issues/detail?id=437549

```

Reported by `bo_xu@foxitsoftware.com` on 2014-12-04 18:20:16

<hr>
 * *Attachment: [community-amenity-contributions-through-rezonings-policy.pdf](https://storage.googleapis.com/google-code-attachments/openjpeg/issue-447/comment-0/community-amenity-contributions-through-rezonings-policy.pdf)*

szukw000:
@mayeut , @detonin , all images of issue #447 (0.jp2 ... 8.jp2) are broken.

All images have an ihdr, a colr and a pclr box inside jp2h, but a cmap box outside jp2h.

KDU refuses to show: "JPX source contains a codestream with a palette (pclr) box, but
no component mapping (cmap) box. This illegal situation ..."

winfried

Issue #235, opened 2013-09-06
CMAP: MTYP == 0 (direct use) not handled properly


gcode-importer:
Originally reported on Google Code with ID 235
```
The attached file appears blurred in openjpeg-2.0 because the final tile part is higher
than expected and so opj_j2k_read_sot exits too early.

I have attached a patch to resolve this.
```

Reported by [slmistry](https://github.com/slmistry) on 2013-09-06 22:04:47

<hr>
 * *Attachment: [blur.patch](https://storage.googleapis.com/google-code-attachments/openjpeg/issue-235/comment-0/blur.patch)*
 * *Attachment: [Bug691816.pdf](https://storage.googleapis.com/google-code-attachments/openjpeg/issue-235/comment-0/Bug691816.pdf)*

detonin:
@mayeut the test you added is failing on myside ... Is this expected?
http://my.cdash.org/viewTest.php?onlydelta&buildid=796186

mayeut:
@detonin,

I've seen this. I forgot to add the dump for issue235.jp2, it's fixed.

PR #534, opened 2015-07-12
Add checks for odd looking cmap & for cmap outside jp2h box

mayeut:
Update uclouvain/openjpeg#235
Update uclouvain/openjpeg#447


mayeut:
@detonin ,

Seems that, as for #254 , there's an encoder out there that generates this kind of data... As those images also contains the issue #254 I guess it's the same encoder...
I don't think we need to make this patch optional (Although it would be possible of course).

If this looks good to you, I'll add data to the test suite before merging this PR

detonin:
@mayeut looks good indeed. thanks for adding data to the test also.

mayeut:
Fixes uclouvain/openjpeg#235
Fixes uclouvain/openjpeg#447

PR #524, opened 2015-06-27
Added OPJ_BIG_ENDIAN code in src/bin/jp2/convert.c for TIFF images with 16-Bit

szukw000:
Added OPJ_BIG_ENDIAN code in src/bin/jp2/convert.c for TIFF images with 16-Bit.

The change in line 1606 seemed to necessary when decompressing the file
'data//input/conformance/file6.jp2': it does not have 16-Bit but 12-Bit. Without
this change the resulting TIFF file was very dark, nearly black.

Tested on Debian PPC64 Jessie 8.1.0.

Issue #322, opened 2014-03-31
limited tif support


gcode-importer:
Originally reported on Google Code with ID 322
```

TIF support is currently provided by libtiff but this has some limitations, in particular
when dealing with non-8 or non-16 bits j2k images.

For instance, file6.jp2 from conformance test set is 12 bits per component, but OpenJPEG
produces a decoded tif file with 16 bits per component. This is wrong : the tiff file
should be 12 bits too.

TIF support should be improved/extended to support this kind of cases.


```

Reported by [detonin](https://github.com/detonin) on 2014-03-31 13:05:19

mayeut:
Closed by PR #535

Issue #264, opened 2014-02-28
convert.c is unmaintainable


gcode-importer:
Originally reported on Google Code with ID 264

```
Just look at convert.c to be convinced. For example have a look at issue #171 and in
particular the patch needed to clamp value returned from openjpeg. All formats are
affected by the issue.

This means that we could have our test suite be valid for PGX output, but invalid for
other, this feel really unconfortable.
```

Reported by [malaterre](https://github.com/malaterre) on 2014-02-28 08:38:42

PR #535, opened 2015-07-16
Update TIFF conversion to support more bit depth.

mayeut:
Update uclouvain/openjpeg#322
Update uclouvain/openjpeg#264

This support 1, 2, 4, 6, 8, 10, 12, 14, 16 bits TIFF both in reading / writing.
Includes (need some testing though) support for BIG ENDIAN (#524)
All tests are OK (well, rounding changed when doing 16 -> 12 bits conversion so need to update non regression data)


szukw000:
@mayeut ,
my tests on PPC64 (resp. LINUX-64) were OK.

BUT: you have forgotten to apply the patch in #518. I had to apply #518. 

The difference between the old code and the patch #518 can be visited by

  http://home.arcor.de/szukw000/cio.txt.zip

For tiftoimage() I have made a small change. Instead setting tBps and tPhoto to 0,
I use the variable 'fails' and show the original values for tBps and tPhoto in case
'fails == 1".

winfried

Issue #459, opened 2014-12-22
OpenJpeg fails to encode components with different precision properly


gcode-importer:
Originally reported on Google Code with ID 459
```
OpenJpeg fails to encode components with different precision properly.

issue203-32x32-bgr16.bmp & issue203-127x64-bgr16.bmp are not encoded properly. The
image color are off, with or without mct.

It seems to happen only when the first component is not the one with highest precision.

bmpmask16toimage in convertbmp.c can easily be modified to produce different combinations.

i.e. to get 5bits precisions on all components (from this bgr565 images) would modify
it as follow :
	bmp_mask_get_shift_and_prec(blueMask,  &blueShift,  &bluePrec);
	bmp_mask_get_shift_and_prec(alphaMask, &alphaShift, &alphaPrec);
	
+	greenPrec -= 1;
+	greenShift += 1;
	
	image->comps[0].bpp = redPrec;
	image->comps[0].prec = redPrec;

```

Reported by [mayeut](https://github.com/mayeut) on 2014-12-22 18:34:33

mayeut:
NR-ENC-issue203-32x32-bgr16.bmp-43-compare_dump2base
NR-ENC-issue203-32x32-bgr16.bmp-43-compare_dec-ref-out2base
NR-ENC-issue203-33x33-bgr16.bmp-44-compare_dump2base
NR-ENC-issue203-33x33-bgr16.bmp-44-compare_dec-ref-out2base
NR-ENC-issue203-127x64-bgr16.bmp-49-compare_dump2base
NR-ENC-issue203-127x64-bgr16.bmp-49-compare_dec-ref-out2base

All those tests are failing because of that (reference data not created because of that issue in fact)

detonin:
Hi thanks for all the work. There is sth more fundamental I don't get here: COC and QCC are used to define different coding style and quantization parameters for a given component (by "different"; I mean "different than what is defined in QCD and COD").
The issue here seems to be at encoding side, which is surprising as a given image does not actually require itself that one of its components be encoded differently, this is more sth related to the options one wants to give at the commandline ... Or am I missing sth ? 
So I would say having the encoder able to write QCD and COC is nice but is this really solving the issue with the files given in #459 or only hiding it in some way ?

mayeut:
@detonin, those images are rgb16: red 5 bits, green 6 bits, blue 5bits.
[...]

PR #551, opened 2015-07-29
Add COC/QCC in main header when needed

mayeut:
Update uclouvain/openjpeg#459


malaterre:
Can we get a little bit more info about this ? Maybe a small unit test that exhibit the issue ?

mayeut:
@malaterre, there's already test cases in the test suite for this as mentioned in #459:
NR-ENC-issue203-32x32-bgr16.bmp-43-compare_dump2base
NR-ENC-issue203-32x32-bgr16.bmp-43-compare_dec-ref-out2base
NR-ENC-issue203-33x33-bgr16.bmp-44-compare_dump2base
NR-ENC-issue203-33x33-bgr16.bmp-44-compare_dec-ref-out2base
NR-ENC-issue203-127x64-bgr16.bmp-49-compare_dump2base
NR-ENC-issue203-127x64-bgr16.bmp-49-compare_dec-ref-out2base

The tests are now ok with this PR

basically what it does:
When when setting up main header writing in `opj_j2k_setup_header_writing`, add new procedures that will write COC/QCC. Those procedures will only write something if COC/QCC is different from COD/QCD so that there's no regression with previous implementation for normal cases.

Issue #488, opened 2015-04-28
Do not link with -ffast-math


gcode-importer:
Originally reported on Google Code with ID 488
```
Linking shared libraries with -ffast-math is dangerous.  It causes the SSE mxcsr flags
to be set on library *load*, which can cause various issues for programs the unsuspectingly
load such a library.  See:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55522
https://bugzilla.redhat.com/show_bug.cgi?id=1127544

for some discussion, examples.

Unfortunately, CMAKE_C_FLAGS are also used during link, so the current:

   # Do not use ffast-math for all build, it would produce incorrect results,
only set for release:
   SET(CMAKE_C_FLAGS_RELEASE "-ffast-math ${CMAKE_C_FLAGS_RELEASE}")

triggers this.

According to http://public.kitware.com/pipermail/cmake/2015-April/060479.html the preferred
solution would be something like:

target_compile_options(openjpeg PRIVATE
  $<$<CONFIG:Release>:-ffast-math>)

The attached patches appears to achieve this (for 1.5.1 and trunk).

Another wrinkle on this is that -DCMAKE_BUILD_TYPE=RelWithDebInfo or MinSizeRel will
also cause -ffast-math not to be used at all.  Not sure if that is an issue or not.

```

Reported by `orion@cora.nwra.com` on 2015-04-28 21:43:08

<hr>
 * *Attachment: [openjpeg-fast-math.patch](https://storage.googleapis.com/google-code-attachments/openjpeg/issue-488/comment-0/openjpeg-fast-math.patch)*
 * *Attachment: [openjpeg2-fast-math.patch](https://storage.googleapis.com/google-code-attachments/openjpeg/issue-488/comment-0/openjpeg2-fast-math.patch)*

rdieter:
I can take care of importing the proposed patches

rdieter:
Hrm, apparently I don't have commit access yet :(

mayeut:
Done on master

mayeut:
This requires CMake 2.8.12
Minimum CMake version was 2.8.2 until now.
Either disable -ffast-math before 2.8.12 or bump minimum CMake version to 2.8.12 ?

mayeut:
@detonin,

Feel free to close if the implemented solution is OK.
> disable -ffast-math before 2.8.12

detonin:
Implemented solution is ok. Thanks. Closing.
[...]

PR #554, opened 2015-07-30
better -ffast-math handling

rdieter:
issue #488

See also
http://public.kitware.com/pipermail/cmake/2015-April/060479.html


mayeut:
@rdieter,

Could you update your PR to include the same guards as in commit c5dd6035c942ab5969b7dc3a873e95e3b542f544 please ?

malaterre:
I do not believe we handle this in openjpeg

mayeut:
We do on master and now on branch openjpeg-1.5 after commit 650df81e09a3362010a4047e04313a24a4a687c8

malaterre:
I believe that openjpeg lossless mode currenly only works with `-fexcess-precision=standard`

mayeut:
Lossless tests are okay without `-fexcess-precision=standard` since #591 and #571 are fixed. Am I missing something ?

malaterre:
That's precisely my point. OpenJPEG lossless code path was actually lossy until a couple of weeks back (project is > 10 yrs old!). So I am suggesting to be a little bit modest and do not pretend we handle those advanced options for now. If distro packager wants to use this flags then it is their responsabilities, that's fine.

mayeut:
I understand your point. However this is another issue than the one taken care of here which used to mess with floating point configuration for people linking with openjpeg.

malaterre:
point taken. sorry for the noise.

PR #553, opened 2015-07-30
better -ffast-math handling

rdieter:
issue #488

See also
http://public.kitware.com/pipermail/cmake/2015-April/060479.html

PR #555, opened 2015-07-30
better -ffast-math handling

rdieter:
issue #488

See also
http://public.kitware.com/pipermail/cmake/2015-April/060479.html


mayeut:
Tested on MacOS 10.10 Xcode 6.4 x86/x64 / Windows 10 vc14 x64 / Ubuntu 14.04 gcc 4.8.4 x64 with no regression in the tests.

mayeut:
I'll let someone else do the testing on 2.1 (#553) & 1.5 (#554)

PR #569, opened 2015-08-23
Updated: opj_flviewer as well as cmyk, esycc and cielab code

szukw000:


mayeut:
Duplicated by #585

szukw000:
@mayeut ,
what does 'Duplicated by #585' mean?

winfried

mayeut:
@szukw000,

It means that you opened PR #585 which does the same thing as this one but is up to date, so this one can be closed.

PR #578, opened 2015-09-08
Travis-ci update

mayeut:


mayeut:
@detonin,

Not sure we can use Kakadu trial binaries for automated builds.
Any idea ?

detonin:
You mean legally ? 
If so, I think we actually can, but we need to add somewhere in the wiki the copyright notice. From the kakadu website:

"You are free to trial these executables and even to re-distribute them, so long as such use or re-distribution is accompanied with this copyright notice and is not for commercial gain. Note: Binaries can only be used for non-commercial purposes."

http://kakadusoftware.com/downloads/

As an open source project, I think we can say that the kakadu binaries are not used for any commercial purpose.

njdoyle:
OpenJPEG's BSD license allows for commercial use and that Kakadu notice does not. Isn't that an incompatibility?

detonin:
OpenJPEG allows for commercial use but is not itself a commercial product. 
Hence, AFAIU the kakadu notice, using kakadu trial binaries in the maintenance of a non-commercial software is ok. 
What would be forbidden is to embed kakadu binaries in a bigger software and sell it. 

I leave open until we have added the kakadu notice in a meaningful place on our GH platform.

njdoyle:
This still conflicts when OpenJPEG is used by someone else commercially because it's distributed with artifacts that have a more restrictive non-commercial license.

mayeut:
@njdoyle,

Kakadu trial binaries are never redistributed with OpenJPEG.
The test suite can use those binaries if present.

@detonin,
I guess we can use it for testing in this repo as you said.
Let me ask the question this way:
If a forked repo of OpenJPEG is held by a company that will use it in a commercial product, can they use kakadu with their travis-ci environment  ? Forked repository are always public so this might not be an issue, however, they can duplicate the repo & make it private.

If yes, then I guess we don't have to do much since terms & conditions are embedded in the downloaded archive.
If they can't, I think it's better to conditionally download/use Kakadu & add the notice before the condition.

I think in any case, the later option can't harm.

detonin:
@mayeut,
Yes, I think conditional download with OPJ_NONCOMMERCIAL is safe. Thanks for this. If companies are using our travis file (with kakadu) to develop commercial products, this is their own responsibility as long as the copyright notice is clearly indictaed in our file. Maybe an improvement of the travis.yml file would be to echo the copyright notice so that it appears in the log as well. Could you update this in that way and merge the PR ?

Later on, an additional wiki page shall describe more in detail all our continuous integration tools and the copyright notice shall be found there as well.

Issue #571, opened 2015-09-01
opj_compress fails to compress lossless on gcc/x86 (-m32)


malaterre:
I am having an issue with OPJ 1.5 and the following file:

http://gdcm.sourceforge.net/thingies/lossless_issue.pgm

It looks as if I cannot do a round-trip lossless compression / decompression, within a 32bits schroot. Everything is fine within a 64bits schroot system (debian jessie if that matter).

malaterre:
Steps from debian/jessie i386:

$ image_to_j2k -i lossless_issue.pgm -o lossless_issue.j2k
$ j2k_to_image -i lossless_issue.j2k -o lossless_issue.opj.pgm
$ vim -b lossless_issue.opj.pgm
-> remove comment
$ vbindiff lossless_issue.pgm lossless_issue.opj.pgm

malaterre:
I can reproduce the exact same behavior using openjpeg 2.1.0 (as packaged in debian).

malaterre:
It seems the compression step is not working as expected:
{{{
$ crc32 *
49d5c244	lossless_issue.pgm
c611132c	lossless_issue.sid32.j2k
a913d7c3	lossless_issue.sid32.kdu.pgm
a913d7c3	lossless_issue.sid32.pgm
9922c6b1	lossless_issue.sid.j2k
49d5c244	lossless_issue.sid.kdu.pgm
49d5c244	lossless_issue.sid.pgm
}}}

malaterre:
The good news is that if I run:
```
$ valgrind  opj_compress -i lossless_issue.pgm -o lossless_issue.sid32.j2k
```

Everything works as expected. So valgrind is doing some kind of initialisation (rounding?) step that is missing is normal code executition.

malaterre:
If I use now clang-3.5 from my sid 32bits chroot everything works nicely. So the issue is not within the libc and such, but rather in the code generated by gcc.

malaterre:
gcc-4.8, gcc-4.9 and gcc-5.2 all have the same symptoms

malaterre:
If I recompile openjpeg with ` -fsanitize=undefined` I get:

/home/mathieu/tmp/opj-bug/openjpeg/src/lib/openjp2/t1.c:1517:28:
runtime error: left shift of negative value -128

ref:

https://github.com/uclouvain/openjpeg/blob/master/src/lib/openjp2/t1.c#L1517

malaterre:
[...]

PR #579, opened 2015-09-09
Correct lossless issue on linux x86

mayeut:
Update uclouvain/openjpeg#571

### DIFF
--- a/src/lib/openjp2/tcd.c
+++ b/src/lib/openjp2/tcd.c
@@ -565,7 +565,7 @@ OPJ_BOOL opj_tcd_rateallocate(  opj_tcd_t *tcd,
                         opj_t2_destroy(t2);
                 } else {
                         success = OPJ_TRUE;
-                        goodthresh = min - DBL_MAX; /* do not rely on float equality for lossless transform */
+                        goodthresh = min - DBL_EPSILON; /* do not rely on float equality for lossless transform */
                 }
 
                 if (!success) {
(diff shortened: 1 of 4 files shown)


malaterre:
this fixes the symptoms, not the bug

mayeut:
@malaterre,

opj_tcd_makelayer takes a threshold that's computed in opj_tcd_rateallocate

the computation for the threshold in case of lossless transform is identical to the computation in  opj_tcd_makelayer. Thus, it expects "computation" == stored_threshold to be true at some point.
The stored threshold will always be rounded to 64bits. "computation" precision will depend on compilation options, architectures, ...

IMHO, this PR removes the dependency on 64bits precision.

If you fill it does not fix the issue properly, I'll revert the "invalid fix" & only merge the tests in.

mayeut:
@malaterre,

Given the definition of DBL_EPSILON:
```c
#define DBL_EPSILON     2.2204460492503131e-016 /* smallest such that 1.0+DBL_EPSILON != 1.0 */
```
It might not be a good fix.

We might want to make the check with epsilon directly in the test in opj_tcd_makelayer
```c
if ((thresh - (dd / dr)) <= DBL_EPSILON) /* do not rely on float equality, check with DBL_EPSILON margin */
	n = passno + 1;
```

I think the later would be OK, do you  agree ?

malaterre:
@mayeut that's a lot better ! I still do not like the equality (does not work for floating point anyway). So I would write:
```
if ((thresh - (dd / dr)) < DBL_EPSILON)
```

Do we need an `fabs` here also ?

mayeut:
@malaterre,

I removed ```=``` from the test. It's now:
```c
if ((thresh - (dd / dr)) < DBL_EPSILON)
```

``` fabs``` would have been needed if original test was ```==```. This is not the case here.

malaterre:
ACK. Everything is fine for me. Thx

Issue #356, opened 2014-06-13
Update 2 to memory management


gcode-importer:
Originally reported on Google Code with ID 356

```
This is the next update to bug 253 (https://code.google.com/p/openjpeg/issues/detail?id=253)
and patches on top of OpenJPEG2.1.

This patch propagates the codec to all areas where memory management is used and also
means that the existing code can be further extended to swap out fprintf statements
for the correct callback message.

The patch has been regression tested using the ghostscript test files.
```

Reported by [slmistry](https://github.com/slmistry) on 2014-06-13 22:52:42

slmistry:
Hi Matthieu,

Any update on when this will be merged in and released?

Shailesh

detonin:
@slmistry 

Next release (due this month) will be 2.1.1 and 1.5.3, meaning that there is no plan to break API nor ABI. So your patch will not make it in the next release, I'm sorry about that. However, we plan to quickly release a 2.2 version where your patch should fit. Sorry for the delay but the patch you propose is huge and breaks both API and ABI so it takes time.

In the mean time, as we switched from GCode to Github, do you think you could fork openjpeg and submit a pull request with the latest version of the patch (I guess this is this one: https://storage.googleapis.com/google-code-attachments/openjpeg/issue-356/comment-22/issue356-r2997.patch) ? This would ease the process of merging your proposal. 

Thanks,

Antonin

slmistry:
Antonin,

I have applied the patch as requested but a number of hunks failed so I am working my way through them manually and it will take some time. Will update you later as things progress.

Shailesh

detonin:
ok thanks

slmistry:
Antonin,

The code is now complete and regression tested so I have submitted a pull request.

Shailesh

Issue #253, opened 2014-01-12
Update to memory management


gcode-importer:
Originally reported on Google Code with ID 253

```
The team at Ghostscript have asked me to contact you about the possibility of updating
the memory management in OpenJPEG2.

In the event of an error condition arising, it can allow the calling app to do a complete
memory cleanup. Another use case is on systems (usually embedded, but not always) where
the system memory manager is poorly optimised for this type of application - being
able to "hook" the memory calls allows a calling app to implement an optimised solution
(for example, the chunk allocator in Ghostscript).

We've had cases where developers wanted Ghostscript to use their special memory manager
calls, which is easy in Ghostscript  but means code changes in OpenJPEG2. A bug for
this has been raised in Ghostscript and can be found at http://bugs.ghostscript.com/show_bug.cgi?id=694691.
The bug report outlines the problems encountered and is hopefully a starting point
for us to discuss a possible solution especially the API changes this would require.

The first step in these changes is to make opj_codec_private_t available through an
include file. Once you are happy with this we can propagate the main context to the
lower function calls.

The attached files have been tested on windows and Linux using OpenJPEG2.
```

Reported by [slmistry](https://github.com/slmistry) on 2014-01-12 15:52:43

<hr>
- _Attachment: [openjpeg2.patch](https://storage.googleapis.com/google-code-attachments/openjpeg/issue-253/comment-0/openjpeg2.patch)_
- _Attachment: [opj_codec.h](https://storage.googleapis.com/google-code-attachments/openjpeg/issue-253/comment-0/opj_codec.h)_

PR #582, opened 2015-09-12
Update 2 to memory management #356

slmistry:
I have applied the issue356-r2997.patch, a number of hunks failed and had to be applied manually. I then corrected a number of errors/warnings and a memory leak in the image destroy functionality.

This code has successfully passed through all the regression files in the Ghostscript repository.


mayeut:
527 errors in OpenJpeg test suite with this PR:
http://my.cdash.org/buildSummary.php?buildid=823368

slmistry:
The last update shows 17 test failures but they appear to be unrelated to this code change. Please can you confirm if this is ok or if you require further input from me.

julienmalik:
See #356

malaterre:
@slmistry could you confirm that you tried to preserved the old API and provided a new one (let's call it the 'manager') ?

slmistry:
The code was checked with/without the manager to ensure it would continue to work as is.

malaterre:
@slmistry did you provide a mechanism to detect case where an application mixes both API, eg:

```
         l_codec = opj_manager_create_compress(l_manager, OPJ_CODEC_JP2);
[...]
         l_codec = opj_create_compress(OPJ_CODEC_JP2);
```

slmistry:
There is no mechanism in place to check for the user mixing API calls but it does seem highly improbable.

If you start with opj_create_compress, then you would not suddenly create an instance of l_manager for your remaining function calls.

Conversely if you start with opj_manager_create_compress, you have already made an instance of l_manager and it should be apparent that you need to keep using it.

malaterre:
So @slmistry confirms that there are two dual (conflicting API) in this patch. I suggest two way to improve this patchet:

* hard path: remove the deprecated API
* soft path: move the deprecated API in between OPJ_USE_MEMORY_MANAGER , update the documentation to mention the deprecated API is only there not to break API (allow simple recompilation for current application).

I do not buy the argument that "people wont mix API" at all, and I do not want OpenJPEG team to maintain a dual API where it is not clear where the benefit of the old one is.

slmistry:
Just for reference, this memory management update has been over 3 years in the making. 

It started on 12/Jan/2014 in issue #253 (https://github.com/uclouvain/openjpeg/issues/253)

It then progressed into issue #356 (https://github.com/uclouvain/openjpeg/issues/356) a period of over a year were it was reworked numerous times exactly as your team requested.

Now nearly 2 years after it was finally accepted onto your milestone list you appear to be requesting even more changes!

slmistry:
Is it possible to get an update on this? I am happy to implement the hard path if you can confirm that it will potentially go into a release fairly soon.

Issue #581, opened 2015-09-11
Use travis-ci matrix build


mayeut:
travis-ci is now configured for basic operation.
Time to improve & go to matrix build.

A few things to keep in mind:
- There's a 50 minutes limit per command on travis-ci so a valgrind analysis is probably not a good idea (at least a full one).
- There's an upload limit of 250MB daily (I think) on CDash. I don't know how much is 1 build... We'll need to limit the matrix size not to flood CDash & be fair with travis-ci which doesn't seem to limit number of elements.

From travis-ci documentation:
> Please take into account that Travis CI is an open source service and we rely on worker boxes provided by the community. So please only specify as big a matrix as you actually need.

So, what's needed ? I'd say:
- linux       gcc x86_64 Release Build
- linux       gcc x86_64 Debug Build
- linux       gcc i386 Release Build
- linux  mingw x86_64 Release Build
- linux  mingw i386 Release Build
- linux    clang x86_64 ASan Debug `-O1` Build
- macos clang x86_64 Release Build

Build options:
- BUILD_CODEC
- BUILD_THIRDPARTY

I think it shall cover the needs.
It also might be time to restore dev-utils folder or equivalent so that travis-ci .travis.yml doesn't become unreadable (it may already be unreadable)

malaterre:
I'd like to see a `gcc -m32` if the debian host are properly multi-arched, maybe replaced the native `gcc Release` with it to keep the matrix small.

malaterre:
BTW, if time is an issue we could have:
* Lots of compilation (matrix with gcc/clang/mingw) but *without* any test
* Keep the full test suite for 2/3 main OS

mayeut:
1. Unfortunately, ```-m32``` is not supported on their linux hosts but 32 bits builds can be built on osx though.
2. The time limit is per matrix-element so there may be no need not to run tests.
3. I don't think mingw is supported out of the box on their hosts

I'll update the first post to keep a proposed matrix there.

malaterre:
Meant to say cross building with mingw, eg:

https://packages.debian.org/jessie/mingw-w64

mayeut:
There are multiple environment on travis-ci.
By default, ```sudo``` is disabled, so no way to install packages.
Need to check how to change environment so that ```sudo``` is supported (this would allow for ```gcc -m32```as well).

mayeut:
travis-ci documentation is misleading... apt is supported (marked BETA in logs though).
[...]

PR #584, opened 2015-09-12
Add Travis-ci build matrix

mayeut:
Update #581


mayeut:
Travis-ci check fails because of leak in master branch.
http://my.cdash.org/viewDynamicAnalysis.php?buildid=823401

Issue #619, opened 2015-10-06
Use of enum variable for bit flags prevents compilation as C++ source


smuehlst:
I'm compiling OpenJPEG with the .NET compiler in C++ mode. There are two problems that prevent successful compilation out of the box. One is missing casts for return values of opj_malloc()/opj_calloc(). For that I already created pull request #618.

The other problem is the use of an enum typedef for a variable that is intended for holding bit flags. From src/lib/openjp2/cio.h:

    typedef enum
    {
            opj_signed_sentinel		= -1, /* do not use in code */
            opj_stream_e_output		= 0x1,
            opj_stream_e_input		= 0x2,
            opj_stream_e_end		= 0x4,
            opj_stream_e_error		= 0x8
    }
    opj_stream_flag ;

    typedef struct opj_stream_private
    {
        ...
        /**
         * Flags to tell the status of the stream.
         */
        opj_stream_flag m_status;
    }
    opj_stream_private_t;

And in src/lib/openjp2/cio.c for example:

    l_stream->m_status |= opj_stream_e_input;

This fails to compile in C++ mode. The declaration of the `m_status` member as `opj_stream_flag` in my view does not add value here, because as the variable can contain a combination of the `opj_stream_*` flags it is not restricted anyway to the actual members of the` opj_stream_flag` enumeration.

To be able to compile in C++ mode I changed the type of the `m_status` variable from `opj_stream_flag` to `OPJ_INT32` in my local repository. Would you consider to apply this change? If so, I would create a corresponding pull request.

mayeut:
@smuehlst,

I guess the flags should all be declared as preprocessor defines:
```c
#define OPJ_STREAM_STATUS_OUTPUT  0x1U
#define OPJ_STREAM_STATUS_INPUT   0x2U
#define OPJ_STREAM_STATUS_END     0x4U
#define OPJ_STREAM_STATUS_ERROR   0x8U
```
 
The structure shall be defined as:
```c
typedef struct opj_stream_private
{
    ...
    /**
     * Flags to tell the status of the stream.
     */
    OPJ_UINT32 m_status;
}
opj_stream_private_t;
```

I think it's best to use unsigned integers for bitfields (some static analyzers will report a warning for bitwise operations on signed integers).

smuehlst:
@mayeut: I see, you are right that an unsigned integer makes more sense. I did not pay attention that the enum member `opj_signed_sentinel` Isn't actually used anywhere.

[...]

PR #621, opened 2015-10-07
Fix code for compilation with C++ compiler

stweil:
It was tested with g++ (and still works with gcc and clang).
Issue #619 is fixed by these patches.


stweil:
The first patch is no longer valid because the enum type was removed.
The second patch still applies and is now available in pull request #623.

PR #622, opened 2015-10-07
Fixed problem that C++ compilation failed because of enum variable.

smuehlst:
Replaced usage of an enum typedef for a variable that is used
for bit flags with OPJ_UINT32 type.


smuehlst:
This change is intended to fix issue #619.

PR #623, opened 2015-10-07
Fix warnings for C++

stweil:
g++ complains about invalid conversions like these ones:

error: invalid conversion from ‘const char*’ to ‘char*’ [-fpermissive]
error: invalid conversion from ‘void*’ to ‘opj_precision* {aka opj_prec*}’ [-fpermissive]


mayeut:
@stweil,
Thanks, I was about to cherry pick the commit with this fix from PR #621 
You made that much easier.

PR #636, opened 2015-10-17
Update allocation functions

mayeut:


malaterre:
@mayeut could you please add some more details on your behavior change for `aligned_realloc` ?

mayeut:
@malaterre, not much to be said.
`alignment` wasn't check (needs to be power of 2)
`a_ptr` wasn't check either (if NULL, a crash would have happen)

The main thing is that the `aligned_realloc` is not used anymore (no need to, an `aligned_free` + `aligned_malloc` is all that's needed).

PR #638, opened 2015-10-18
Fix format strings and unneeded assignment

stweil:
The static code analyzer cppcheck warns about unsigned integers
which use "%d" in the format string.

It also warns about an unneeded assignment.

Fix both issues.


mayeut:
I'm surprised to see these are the only places where format specifier is "wrong".

Issue #640, opened 2015-10-19
HP compiler warns about redeclaration of static function


smuehlst:
The HP-UX compiler issues the following warning when compiling source file src/lib/openjp2/dwt.c:

    cc: "dwt.c", line 798: warning 562: Redeclaration of "opj_v4dwt_decode" with a different storage class specifier: "opj_v4dwt_decode" will have internal linkage.

The forward declaration looks like this:

    static void opj_v4dwt_decode(opj_v4dwt_t* restrict dwt);

While the actual function signature looks like this:

    void opj_v4dwt_decode(opj_v4dwt_t* restrict dwt)

The actual function should also be declated as `static`.

smuehlst:
Same problem in source file src/lib/openjp2/t2.c:

cc: "t2.c", line 1341: warning 562: Redeclaration of "opj_t2_init_seg" with a different storage class specifier: "opj_t2_init_seg" will have internal linkage.

Issue #243, opened 2013-10-08
OpenJPEG doesn't compile on mac with gcc 4


gcode-importer:
Originally reported on Google Code with ID 243
```
Several methods are forward declared as static but the actual definition does not have
the static keyword (gcc4 with our build options chokes on this).

This was reported in OpenJPEG 2.0.0
MacOSX 10.7.5
Xcode 3.2.6

Attached patch is generated with diff.


```

Reported by `brta.enfocus` on 2013-10-08 12:29:48

<hr>
 * *Attachment: [fix gcc 4 compile errors](https://storage.googleapis.com/google-code-attachments/openjpeg/issue-243/comment-0/fix gcc 4 compile errors)*

mayeut:
Should be fixed but I can't verify this with my compiler.

PR #641, opened 2015-10-19
Fix HP compiler warning about redeclaration of function (#640)

smuehlst:
HP compiler warns:
cc: "dwt.c", line 798: warning 562: Redeclaration of "opj_v4dwt_decode"
with a different storage class specifier: "opj_v4dwt_decode" will have
internal linkage.
cc: "t2.c", line 1341: warning 562: Redeclaration of "opj_t2_init_seg"
with a different storage class specifier: "opj_t2_init_seg" will have
internal linkage.

Fix for issue #640.


mayeut:
@smuehlst, thanks.
I didn't find a way to properly test this with compilers at my disposal when working on #243

Issue #642, opened 2015-10-20
Compilation fails without platform-supplied aligned malloc


smuehlst:
When none of none of `posix_memalign()`, `memalign()` and `_aligned_malloc()` is available, compilation of the current HEAD revision 57996722371d07e9626a36330c3a7a453475d4fd fails.

Discussion is on the mailing list: [Building OpenJPEG without a platform-supplied aligned memory allocation function](https://groups.google.com/forum/#!topic/openjpeg/LHwfVV_mSIg).

mayeut:
@smuehlst, I pushed a version that shall be ok. Tested on my side by forcing this definitions to be used.
Please close the issue if everything works on your side.

smuehlst:
@mayeut: Thanks for the comments and for the update to my implementation. It works fine for me. The only issue is that there's a compiler warning about pointer arithmetic with void * pointers.

I created pull request #644 with a fix for the warning.

mayeut:
@smuehlst, I just merged your PR. Should be ok without warnings now.

smuehlst:
@mayeut: Thanks, all is good now, closing this issue.

PR #643, opened 2015-10-20
Generic aligned malloc implementation.

smuehlst:
Used for platforms where none of posix_memalign(), memalign() and
_aligned_malloc() is available.

Fix for #642.

### DIFF
--- a/src/lib/openjp2/opj_malloc.c
+++ b/src/lib/openjp2/opj_malloc.c
@@ -32,6 +32,10 @@
 #define OPJ_SKIP_POISON
 #include "opj_includes.h"
 
+#ifndef SIZE_MAX
+#define SIZE_MAX ((size_t) -1)
+#endif
+
 static INLINE void *opj_aligned_alloc_n(size_t alignment, size_t size)
 {
   void* ptr;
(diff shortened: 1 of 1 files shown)


mayeut:
@smuehlst, `ptrdiff_t` can't be used here because it might not hold the whole pointer (i.e. `sizeof(ptrdiff_t) <= sizeof(intptr_t)`, no strict equality, this depends on the platform).
I'll try to get something merged in based on your PR but modified with this in mind.

smuehlst:
@mayeut: I used ptrdiff_t because of the following [recommendation on the mailing list](https://groups.google.com/d/msg/openjpeg/LHwfVV_mSIg/wGYHbeDWAgAJ) by @malaterre:

>I do not believe you need the full range of uintptr_t for the mask,  since align is typically stored on unsigned char (8, 16, 32...)  As for the pointer arithmetic, you can simply use ptrdiff_t.

intptr_t and uintptr_t is not available in C89.

malaterre:
I would follow the suggestion from [here](http://stackoverflow.com/a/20495779/136285). Simply use:

#if PTRDIFF_MAX < SIZE_MAX/2
# error "we need an architecture with sufficiently wide ptrdiff_t"
#endif

mayeut:
Why not use this ? Should work on all platforms if I'm not mistaken.
```c
  /*
   * Generic aligned malloc implementation.
   * Uses ptrdiff_t for the integer manipulation of the pointer, as
   * uintptr_t is not available in C89.
   */
  alignment--;
  {
    size_t offset;
    OPJ_UINT8 *mem;

    /* Room for padding and extra pointer stored in front of allocated area */
    size_t overhead = alignment + sizeof(void *);
		
    /* let's be extra careful */
    assert(alignment <= (SIZE_MAX - sizeof(void *)));

    /* Avoid integer overflow */
    if (size > (SIZE_MAX - overhead)) {
      return NULL;
    }

    mem = (OPJ_UINT8*)malloc(size + overhead);
    if (mem == NULL) {
      return mem;
    }
    /* offset = ((alignment + 1U) - ((size_t)(mem + sizeof(void*)) & alignment)) & alignment; */
    /* Use the fact that alignment + 1U is a power of 2 */
    offset = ((alignment ^ ((size_t)(mem + sizeof(void*)) & alignment)) + 1U) & alignment;
    ptr = (void *)(mem + sizeof(void*) + offset);
    ((void**) ptr)[-1] = mem;
  }
```

mayeut:
I'd like to point out that the implementation of the `realloc` assumes that alignment requirement doesn't change between the `malloc` and `realloc`.
I think this is fine but should be noted. According to MSDN, `_aligned_realloc` has the same limitation anyway.
https://msdn.microsoft.com/fr-fr/library/y69db7sx.aspx

mayeut:
Merged in d48be27f5aa018d398cf9369b2c664bee4ce6e19

PR #644, opened 2015-10-21
Avoid pointer arithmetic with (void *) pointers

smuehlst:
Taking the difference of two void * pointers is not defined according to the C specification, and for example the z/OS compiler warns:

./opj_malloc.c:169   Operation between types `void*` and `void*` is not allowed.


mayeut:
Exact, I forgot about this & got no warning on my compiler... Difference is only valid for 2 non void* pointers of the same type.

PR #648, opened 2015-10-29
Fix support of posix_memalloc for Linux

stweil:
posix_memalloc is only declared conditionally in stdlib.h,
so add one of the possible definitions to get the declaration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -231,8 +231,10 @@ check_include_files(malloc.h OPJ_HAVE_MALLOC_H)
 include(CheckSymbolExists)
 # _aligned_alloc https://msdn.microsoft.com/en-us/library/8z34s9c6.aspx
 check_symbol_exists(_aligned_malloc malloc.h OPJ_HAVE__ALIGNED_MALLOC)
-# posix_memalign
+# posix_memalign (needs _POSIX_C_SOURCE >= 200112L on Linux)
+set(CMAKE_REQUIRED_DEFINITIONS -D_POSIX_C_SOURCE=200112L)
 check_symbol_exists(posix_memalign stdlib.h OPJ_HAVE_POSIX_MEMALIGN)
+unset(CMAKE_REQUIRED_DEFINITIONS)
 # memalign (obsolete)
 check_symbol_exists(memalign malloc.h OPJ_HAVE_MEMALIGN)
 #-----------------------------------------------------------------------------
--- a/src/lib/openjp2/opj_config_private.h.cmake.in
+++ b/src/lib/openjp2/opj_config_private.h.cmake.in
@@ -28,6 +28,13 @@
 /* check if function `posix_memalign` exists */
 #cmakedefine OPJ_HAVE_POSIX_MEMALIGN
 
+#if !defined(_POSIX_C_SOURCE)
+#if defined(OPJ_HAVE_FSEEKO) || defined(OPJ_HAVE_POSIX_MEMALIGN)
+/* Get declarations of fseeko, ftello, posix_memalign. */
+#define _POSIX_C_SOURCE 200112L
+#endif
+#endif
+
 /* Byte order.  */
 /* All compilers that support Mac OS X define either __BIG_ENDIAN__ or
 __LITTLE_ENDIAN__ to match the endianness of the architecture being


src/lib/openjp2/opj_includes.h
```diff
@@ -44,6 +44,10 @@
    Standard includes used by the library
  ==========================================================
 */
+
+/* Get declarations of fseeko, ftello, posix_memalign. */
+#define _POSIX_C_SOURCE 200112L
```
malaterre: opj_includes.h gets installed. So we are polluting everyone else code with our setting. Can't this be moved to say opj_config_private.h.cmake.in ?

CMakeLists.txt
```diff
@@ -225,6 +225,8 @@ CHECK_INCLUDE_FILE("unistd.h"       HAVE_UNISTD_H)
 include(TestLargeFiles)
 OPJ_TEST_LARGE_FILES(OPJ_HAVE_LARGEFILES)
 
+set(CMAKE_REQUIRED_DEFINITIONS -D_POSIX_C_SOURCE=200112L)
```
stweil: I'm not an cmake expert. How can I set that definition only for the posix_memalloc test?

src/lib/openjp2/opj_includes.h
stweil: Ok, I'll try an alternative solution.

CMakeLists.txt
malaterre: I would use `unset` when done

stweil:
I updated the patch (and also fixed the commit message).

stweil:
Patch updated because of commit 38ffbbe42ee36cf2ab402977cfd5dda471355052.

mayeut:
@stweil,
This PR fails to build under linux & macos. c.f. http://my.cdash.org/index.php?project=OPENJPEG&date=2015-11-01

stweil:
@mayeut: It should be fixed now. The failure only occurred when building thirdparty code, too.

stweil:
Next iteration: Windows does not provide unistd.h.

stweil:
Ping? I think this request can be pulled.

mayeut:
@stweil seems that reverting the fix for tiff broke the build. Can you merge the new master branch in to check that all tests are OK ?

stweil:
@mayeut, obviously the rebased patch passes all checks.

malaterre:
That's really bad, the official reference is actually missing from http://pubs.opengroup.org/onlinepubs/009695399/functions/posix_memalign.html. Is the Linux implementation a bit too restrictive ?

malaterre:
I am ready to merge thie PR until I understand the comment about `feeko/ftello`. We had a mechanism where `_FILE_OFFSET_BITS == 64` was set. Isn't it the case anymore ?

stweil:
fseeko, ftello have the same requirements as posix_memalign. In the current git master, the compiler complains about missing declarations for those functions. The patch fixes that problem, too.

malaterre:
@stweil that is stlighty incorrect. `fseeko/ftello` are available as soon as you define `_FILE_OFFSET_BITS == 64`, hence my request for more comment on your side. Does your compiler requires `_POSIX_C_SOURCE` ?

stweil:
On my 64 bit Debian GNU Linux, _FILE_OFFSET_BITS is not set (in src/lib/openjp2/opj_config_private.h). Therefore I get these compiler messages:

    [  1%] Building C object src/lib/openjp2/CMakeFiles/openjp2.dir/openjpeg.c.o
    /openjpeg/src/lib/openjp2/openjpeg.c: In function ‘opj_get_data_length_from_file’:
    /openjpeg/src/lib/openjp2/openjpeg.c:101:2: warning: implicit declaration of function ‘fseeko’ [-Wimplicit-function-declaration]
      OPJ_FSEEK(p_file, 0, SEEK_END);
      ^
    /openjpeg/src/lib/openjp2/openjpeg.c:102:2: warning: implicit declaration of function ‘ftello’ [-Wimplicit-function-declaration]
      file_length = (OPJ_OFF_T)OPJ_FTELL(p_file);
      ^

Extract from the man page of fseeko:

    Feature Test Macro Requirements for glibc (see feature_test_macros(7)):

       fseeko(), ftello():
           _FILE_OFFSET_BITS == 64 || _POSIX_C_SOURCE >= 200112L || _XOPEN_SOURCE >= 600
           (defining the obsolete _LARGEFILE_SOURCE macro also works)

So defining one of _FILE_OFFSET_BITS, _POSIX_C_SOURCE or _XOPEN_SOURCE fixes the compiler warnings. Maybe _FILE_OFFSET_BITS is set on 32 bit Linux. Then the definition of _POSIX_C_SOURCE is not needed for fseeko, but it won't harm.

detonin:
@stweil Thanks. Merged.

Issue #658, opened 2015-11-17
"master" does not build on ubuntu


butlnor:
When building openjpeg's sources from the "master" branch on Ubuntu, it fails. 
I have Ubuntu 14.04, with devel tools, libboost, CURL and fcgi installed.

It fails when linking "opj_server", as no reference to the memory functions, such as "opj_malloc" or "opj_free", is found.

However, manually replacing opj_* functions with "malloc" and "free" doesn't help, as those are marked as "poisoned" in "opj_malloc.h", and even if this #pragma is commented out, and the build succeeded, running "opj_server" then fails with segmentation fault.

I am appending here the complete original console log, when no openjpeg files were being changed:

============================

user@machine:~/openjpeg-master$ ls

appveyor.yml  AUTHORS  cmake  CMakeLists.txt  CTestConfig.cmake  doc  INSTALL  LICENSE  NEWS  README.md  src  tests  THANKS  thirdparty  tools  wrapping

user@machine:~/openjpeg-master$ cmake . -DBUILD_JPIP:bool=on -DBUILD_JPIP_SERVER:bool=on

-- The C compiler identification is GNU 4.8.4
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check if the system is big endian
-- Searching 16 bit integer
-- Looking for sys/types.h
-- Looking for sys/types.h - found
-- Looking for stdint.h
-- Looking for stdint.h - found
-- Looking for stddef.h
-- Looking for stddef.h - found
-- Check size of unsigned short
-- Check size of unsigned short - done
-- Using unsigned short
-- Check if the system is big endian - little endian
-- Looking for string.h
-- Looking for string.h - found
-- Looking for memory.h
-- Looking for memory.h - found
-- Looking for stdlib.h
-- Looking for stdlib.h - found
-- Looking for stdio.h
-- Looking for stdio.h - found
-- Looking for math.h
-- Looking for math.h - found
-- Looking for float.h
-- Looking for float.h - found
-- Looking for time.h
-- Looking for time.h - found
-- Looking for stdarg.h
-- Looking for stdarg.h - found
-- Looking for ctype.h
-- Looking for ctype.h - found
-- Looking for assert.h
-- Looking for assert.h - found
-- Looking for stdint.h
-- Looking for stdint.h - found
-- Looking for inttypes.h
[...]

PR #659, opened 2015-11-17
Add missing source for the JPIP library and executables (issue #658)

stweil:
They all need opj_malloc and other functions from opc_malloc.c.


stweil:
Ping? This patch waits to be pulled.
Joyeux Noël, Stefan

malaterre:
@mayeut the actual real issue is that this got unoticed. I'd like to keep this open for now.

stweil:
PR #692 fixes issue #691. I suggest to apply PR #659 first, then #692 (reversed order would break checks first and fix them later).

stweil:
@malaterre, @mayeut, what about my suggestion? Can we proceed like that, or do you see more open related issues?

malaterre:
Ok for me !

stweil:
@mayeut, I'd appreciate if I could get this removed from my to-do list. Could you please pull both requests?

PR #666, opened 2015-11-28
T2 optimize

boxerab:
This branch contains the following improvements:

My OpenMP patch
Carl Hetherington's T1 patch
My T2 improvements

I have run the test suite against the patch, and there is no change from the master: same number of tests pass.

The rate control improvements, though very simple, give me a large increase in encoding speed when encoding cinema images : about a 25% speed improvement.  

Net result is: on my 3 year old i7, I can encode 2K frames at around 6.5 frames per second, and decode at around 10 frames per second.


boxerab:
I will try breaking this up into smaller groups of commits, to see what is causing the tests to fail.

stweil:
Did you change indentation from using spaces to using tabs? I suggest to keep spaces.

PR #667, opened 2015-11-28
T2

boxerab:
A few small changes to T2 encoding, which give a big performance improvement.


boxerab:
AppVeyor is in suspended animation. Will close and re-open.

PR #677, opened 2015-12-30
Memory mapped file support

boxerab:
I've added support for memory mapped files, windows and unix..
This will be useful for large images that do not fit into memory, but where one wants to decode a tile.
Currently there is an artificial limit on file size in OpenJPEG, this may help to remove this limit.

Also, with these enhancements, there is potential for dramatic memory savings: for decode, instead
of allocating all of the code block buffers in advance, the block can store a virtual memory pointer from the mapping for all code block segments. Only when the code block is about to be decoded do we need to allocate contiguous memory for the block.

Also for decode, there is no need to keep a separate memory buffer in the stream, and copy into this buffer. The OS will handle buffering.

### DIFF
--- a/src/lib/openjp2/openjpeg.c
+++ b/src/lib/openjp2/openjpeg.c
@@ -33,7 +33,19 @@
 
 #ifdef _WIN32
 #include <windows.h>
-#endif /* _WIN32 */
+#else /* _WIN32 */
+
+#include <errno.h>
+
+#include <stdarg.h>
+#include <stdlib.h>
+#include <sys/stat.h>
+# include <unistd.h>
+# include <sys/mman.h>
+
+#endif
+
+#include <fcntl.h>
 
 #include "opj_includes.h"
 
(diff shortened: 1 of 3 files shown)


src/lib/openjp2/openjpeg.c
```diff
@@ -958,3 +1057,290 @@ opj_stream_t* OPJ_CALLCONV opj_stream_create_file_stream (
...
+			m |= O_TRUNC;
+		break;
+	default:
+		break;
+	}
+	return m;
+}
+
+
+#ifdef WIN32
```
stweil: I suggest using _WIN32 (as in the rest of the file).

src/lib/openjp2/openjpeg.c
```diff
@@ -33,7 +33,30 @@
...
+
+#else /* _WIN32 */
+
+#include <errno.h>
+
+#include <stdarg.h>
+#include <stdlib.h>
+#include <sys/stat.h>
+
+#ifdef HAVE_UNISTD_H
```
stweil: The test for this macro must be enabled in src/lib/openjp2/opj_config_private.h.cmake.in.

src/lib/openjp2/openjpeg.c
```diff
@@ -33,7 +33,30 @@
...
+# include <unistd.h>
+#endif
+
+#ifdef HAVE_FCNTL_H
```
stweil: Neither HAVE_FCNTL_H nor HAVE_IO_H will work without a corresponding CMake test.
There is an example in thirdparty/libtiff/tif_config.h.cmake.in.

boxerab:
Thanks, Stefan. I've tried to address the issues you mentioned.
Do you have time to test this on a unix system?

malaterre:
@boxerab travis-ci is setup to test your branch on *nix system

boxerab:
@malaterre thanks, yes there are errors. should be easy to fix when I have time

boxerab:
*nix should now compile with latest changes.

stweil:
@boxerab, yes, compilation on Debian now works (with some warnings which are addressed in pull request #648).

mayeut:
@boxerab,

x64 builds (linux, windows vc14, macos) are failing under ci because  of regressions:
```
NR-DEC-issue254.jp2-65-decode
NR-DEC-issue254.jp2-65-decode-md5
```
x86 builds segfaults on many tests
windows vc10 does not build: http://my.cdash.org/viewBuildError.php?buildid=883028, related to those warnings: http://my.cdash.org/viewBuildError.php?type=1&buildid=882996

You can find all tests results on [CDash](http://my.cdash.org/index.php?project=OPENJPEG) in the Experimental section.

boxerab:
@malaterre how do the cool kids deal with failed tests?  How to find the error ?

malaterre:
I would suggest having a local VM on your windows host. Eg: http://ruby-journal.com/debug-your-failed-test-in-travis-ci/

boxerab:
Thanks. Test fixed.

boxerab:
Closing this pull request, as my Zero Copy PR includes these changes, plus more improvements.

PR #686, opened 2016-01-06
[libtiff] Add missing include statement for ssize_t

mayeut:


stweil:
See also #681.

mayeut:
@stweil yes, but I'd rather include a file that is known to exist rather than presume that it exists because not on Windows.
Either way, will be fixed.

stweil:
Then I'll have to fix #648, too. unistd.h is a standard header file required by POSIX (see https://en.wikipedia.org/wiki/Unistd.h). I know only one vendor which fails to support it.

stweil:
#648 included #681. I fixed that, so now #648 is now a single commit.

PR #687, opened 2016-01-06
Fix fatal crash on 64 bit Linux

stweil:


stweil:
I did not notice this bug earlier because I normally work with patch #648 applied (so I use posix_memalign with the correct header files instead of memalign with missing header files). I wonder why not more people running 64 bit Linux were affected by the problem.

stweil:
The bug was added recently (in commit 0dc4914b4e5b93ab1d765a541fb342835ada51b5).

Issue #239, opened 2013-09-10
Wrong encoding of small tiles with high level number


gcode-importer:
Originally reported on Google Code with ID 239

```
What steps will reproduce the problem?
1. Take a large image (say 3840x2160) and define the tile size a bit smaller to make
one large tile and three smaller (say -t 3810,2130) and opj_compress with -n 6 or higher.
2. When DWT division comes to input width=1 (OPJ_INT32 rw in dwt.c, line 390) or  the
output for the tile is visibly incorrect.
3. There is a similar problem when tile height comes to 1 and then there is another
DWT level so it repeats the computation with height 1 (it should break in that case).


What is the expected output? What do you see instead?
The output j2k image is visibly malformed on the small tiles. 

What version of the product are you using? On what operating system?
r2345, Ubuntu Linux, 3.2.0-48-generic, x86_64

Please provide any additional information below.


```

Reported by `The.Bobas` on 2013-09-10 08:00:50

rouault:
Fixed per 73d1510d473b7dcfccfdee57e0e511e6791d5091

PR #690, opened 2016-01-08
[git/2.1 regression] Fix opj_write_tile() failure when numresolutions=1

rouault:
When trying the GDAL OpenJPEG driver against openjpeg current master HEAD,
I get failures when trying to create .jp2 files. The driver uses
opj_write_tile() and in some tests numresolutions = 1.

In openjp2/dwt.c:410, l_data_size = opj_dwt_max_resolution( tilec->resolutions,tilec->numresolutions) * (OPJ_UINT32)sizeof(OPJ_INT32);
is called and returns l_data_size = 0. Now in git opj_malloc() has a special case
for 0 to return a NULL pointer whereas previously it relied on system malloc(),
which in my case didn't return NULL.

So only test the pointer value if l_data_size != 0. This makes the GDAL
autotest suite to pass again.

### DIFF
--- a/src/lib/openjp2/dwt.c
+++ b/src/lib/openjp2/dwt.c
@@ -409,7 +409,7 @@ static INLINE OPJ_BOOL opj_dwt_encode_procedure(opj_tcd_tilecomp_t * tilec,void 
 
 	l_data_size = opj_dwt_max_resolution( tilec->resolutions,tilec->numresolutions) * (OPJ_UINT32)sizeof(OPJ_INT32);
 	bj = (OPJ_INT32*)opj_malloc((size_t)l_data_size);
-	if (! bj) {
+	if (l_data_size != 0 && ! bj) {
 		return OPJ_FALSE;
 	}
 	i = l;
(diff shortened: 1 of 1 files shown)


stweil:
Maybe it would be better to return a pointer to a fixed address which points to const memory instead of returning NULL. Then callers would not need special handling.

rouault:
> Maybe it would be better to return a pointer to a fixed address which points to const memory instead of returning NULL. Then callers would not need special handling.

Well you'd have also to make a special case in opj_free() to not free this address. Another solution would be to do malloc(1) when opj_malloc(0) is called

stweil:
Sure, opj_free must ignore that address. Using malloc would only be an option if it had the same defined behavior on all platforms when called with 0.

rouault:
I suggested using malloc(1)

mayeut:
Isn't that linked to #239 ? I think that there's something wrong other than malloc returning `NULL` for size 0 which is, I think, sane behavior.

rouault:
@mayeut I don't think it's connected with #239. Whe comparing HEAD with 2.1, both return l_data_size = 0 but the difference is that HEAD now fails because malloc(0) now returns NULL which is considered as an error.

malaterre:
Could you please handle the case `l_data_size == 0 && bj == NULL`

rouault:
@malaterre Not sure to understand what you mean. It is the nominal case handled by the code after the if (l_data_size != 0 && ! bj)  check. What I know is that with my patch things work again as in 2.1

rouault:
Hum, or perhaps do you mean that if tilec->numresolutions == 1 we could do an early exit return OPJ_TRUE at the beginning of the function ? (since when looking closer the code after the if (l_data_size != 0 && ! bj) check ends up not entering the while (i--) loop )

malaterre:
@rouault my bad. Could you please add comment stating that `bj` will not be used when `l_data_size == 0`.

rouault:
Comment added per 6a1974d40d95d857377a1eb5f4db0d9e24d0ad0a

Issue #676, opened 2015-12-29
Decoding at the precinct level


boxerab:
It would be useful to be able to decode a sub-window inside a tile. This is feedback I got
from the GDAL news group.

This would involve decoding all precincts that overlap with the sub-window, 
including code blocks adjacent to precincts that share a boundary with the 
window. Then, inverse MQ on all blocks, dequantize, sparse inverse DWT on
only the sub window, then inverse MCT on only the sub window.

PR #694, opened 2016-01-11
Decode Sub-region

### LINKED ISSUE #693: -d flag is silently ignored when decoding a single tile
Library always decodes the entire tile, no matter what is set with `-d`

boxerab:
Building on the omnibus branch, this branch adds fast sub-tile decode.

If a decode region is specified using the `-d` parameter, all code blocks that do not contribute
to this region will be ignored, boosting decode speed considerably.

This is useful for decoding a sub region inside of a large single tile image.

Fixes #693 and #676


boxerab:
Strange, all builds pass (ignore ABI test) except gcc 64 bit release, where a lot of tests fail.
Has anyone experienced this, and if so how to trouble shoot ?

mayeut:
@boxerab, could it be related to this warning: 
`openjpeg/src/lib/openjp2/region_mgr.c:194:1: warning: control reaches end of non-void function [-Wreturn-type]`

boxerab:
@mayeut  nice catch - I left some debug code in one of my methods. Thanks.

boxerab:
Well, just 8 broken tests to fix, and they are all related to a single image. Question: do we have any tests in the suite that test the `-d` flag : setting a decode region ?  Also, could we add a test to test decoding a region from a specific tile?  i.e.   `-d -t 0 `

mayeut:
There are tests using the `-d` option. Check in the file `tests/nonregression/test_suite.ctest.in`and look for `-d`. In fact, I think the newly broken tests are those using this option (file `p0_04.j2k` & `p1_04.j2k`).

boxerab:
@mayeut great, thanks!

src/bin/jp2/opj_decompress.c
```diff
@@ -1420,157 +1468,168 @@ int main(int argc, char **argv)
...
+
+	t_cumulative = opj_clock() - t_cumulative;
 	destroy_parameters(&parameters);
-	if (numDecompressedImages) {
-		fprintf(stdout, "decode time: %d ms\n", (int)( (tCumulative * 1000.0) / (OPJ_FLOAT64)numDecompressedImages));
+
+	if (num_decompressed_images) {
+		fprintf(stdout, "decode time: %d ms \n", (int)( (t_cumulative * 1000) / num_decompressed_images));
 	}
+	getchar();
```
mayeut: @boxerab Wouldn't that be the source of the stalled tests under travis ?

boxerab:
@mayeut  once again, nice catch! I checked in some debug code again.
Thanks.

boxerab:
I tested decoding of a 1000x1000 region from a large 22000x29000 image, tiled with 1024x1024 tiles.

This branch took 600 ms, while the omnibus branch took around 1100 ms, on my system.

Issue #695, opened 2016-01-14
MQ Coder encode: Conditional jump or move depends on uninitialised value(s)


boxerab:
http://my.cdash.org/viewDynamicAnalysisFile.php?id=3409924

      if (*mqc->bp == 0xff) {
		mqc->bp++;
		*mqc->bp = (OPJ_BYTE)(mqc->c >> 20);
		mqc->c &= 0xfffff;
		mqc->ct = 7;
	} else {

This is because mq->bp is initialized to one less than the actual data start address.
It could be possible for uninitialized memory to be equal to 0xFF, in which case there will be an error in the code stream. For lossless encoding, this could lead to lossy encoding.

An easy fix is to check if `mqc->bp < mqc->start` before entering the if statement.
On a modern CPU with branch prediction, there should be no performance hit.

mayeut:
Is that the same than #612 ?

boxerab:
yes, it looks like 612

mayeut:
@boxerab closing this one as duplicate.

Follow-up in #612

Issue #612, opened 2015-09-29
Test 481 reports error in valgrind memcheck


stweil:
Running the test manually with a debug version of opj_compress built from latest git code:

```
/usr/bin/valgrind "--tool=memcheck" --leak-check=full --show-reachable=yes --verbose --track-origins=yes --num-callers=50 bin/opj_compress -i /openjpeg-data/input/nonregression/Bretagne2.ppm -o /tmp/Bretagne2_3.j2k -EPH -M 38
```

Result:

```
==11019== Conditional jump or move depends on uninitialised value(s)
==11019==    at 0x4E5EAB5: opj_mqc_byteout (mqc.c:219)
==11019==    by 0x4E5EBAB: opj_mqc_renorme (mqc.c:241)
==11019==    by 0x4E5ED47: opj_mqc_codelps (mqc.c:269)
==11019==    by 0x4E5F196: opj_mqc_encode (mqc.c:408)
==11019==    by 0x4E67BA4: opj_t1_enc_clnpass_step (t1.c:853)
==11019==    by 0x4E6838A: opj_t1_enc_clnpass (t1.c:991)
==11019==    by 0x4E69FB9: opj_t1_encode_cblk (t1.c:1617)
==11019==    by 0x4E69CDF: opj_t1_encode_cblks (t1.c:1536)
==11019==    by 0x4E71B96: opj_tcd_t1_encode (tcd.c:2038)
==11019==    by 0x4E704CA: opj_tcd_encode_tile (tcd.c:1226)
==11019==    by 0x4E47608: opj_j2k_write_sod (j2k.c:4296)
==11019==    by 0x4E569E9: opj_j2k_write_first_tile_part (j2k.c:10472)
==11019==    by 0x4E562A4: opj_j2k_post_write_tile (j2k.c:10258)
==11019==    by 0x4E55A21: opj_j2k_encode (j2k.c:10009)
==11019==    by 0x4E604E3: opj_encode (openjpeg.c:737)
==11019==    by 0x407183: main (opj_compress.c:1864)
==11019==  Uninitialised value was created by a heap allocation
==11019==    at 0x4C28C20: malloc (vg_replace_malloc.c:296)
==11019==    by 0x4E6FFAD: opj_tcd_code_block_enc_allocate_data (tcd.c:1083)
==11019==    by 0x4E6FCC5: opj_tcd_init_tile (tcd.c:1008)
[...]

PR #697, opened 2016-01-14
Fix potential MQ access of uninitialized data

boxerab:
See issue #695

### DIFF
--- a/src/lib/openjp2/mqc.c
+++ b/src/lib/openjp2/mqc.c
@@ -203,7 +203,7 @@ static opj_mqc_state_t mqc_states[47 * 2] = {
 */
 
 static void opj_mqc_byteout(opj_mqc_t *mqc) {
-	if (*mqc->bp == 0xff) {
+	if ( (mqc->bp >= mqc->start) & (*mqc->bp == 0xff)) {
 		mqc->bp++;
 		*mqc->bp = (OPJ_BYTE)(mqc->c >> 20);
 		mqc->c &= 0xfffff;


mayeut:
See #612

src/lib/openjp2/mqc.c
```diff
@@ -203,7 +203,7 @@ static opj_mqc_state_t mqc_states[47 * 2] = {
 */
 
 static void opj_mqc_byteout(opj_mqc_t *mqc) {
-	if (*mqc->bp == 0xff) {
+	if ( (mqc->bp >= mqc->start) & (*mqc->bp == 0xff)) {
```
mayeut: should probably be `&&` instead of just one `&`.

stweil: Yes, && is needed here. I also suggest removing the space after '('.

stweil:
Why can mqc->bp be lower than mqc->start? Should this problem be handled elsewhere?

src/lib/openjp2/mqc.c
boxerab: No, & is superior because it avoids a branch in the code - C has a short circuit on logical operators so there would be two branches with &&

boxerab:
The way the code works is that `mqc->bp` begins at `mqc->start -1`

stweil:
What about re-ordering the two conditions? I think the (*mqc->bp == 0xff) case occurs less often than the (mqc->bp >= mqc->start), so in most cases we would have one compare operation + one taken conditional branch if we use `&&`.

if ((mqc->bp >= mqc->start) & (*mqc->bp == 0xff)) {

1. compare
2. compare
3. boolean operation
4. conditional branch (mostly branching)

if ((*mqc->bp == 0xff) && (mqc->bp >= mqc->start)) {

1. compare
2. conditional branch (mostly branching)
(if branch was not taken => fast)
3. compare
4. conditional branch (mostly not branching => fast)

boxerab:
Are you familiar with the phrase "re-arranging the deck chairs on the Titanic" ? The whole MQ coder routine needs to be refactored with performance in mind.

stweil:
@boxerab, that's why I love working on free software: I always learn new interesting thinks. I did not know the phrase before, but I see what you mean.

mayeut:
Anyway, this doesn't seem to solve the issue for this test: http://my.cdash.org/viewDynamicAnalysisFile.php?id=3435467

boxerab:
Re-opening this shortly with a proper fix (I hope)

PR #705, opened 2016-01-24
Fix implementation of opj_calloc

stweil:
The parameter "size" is typically the size of some data type and should
never be 0. It is much more important to check the "num" parameter if
defined behaviour is required.

Remove also a comment which is not helpful.


stweil:
Ping?

stweil:
Ping²?

Issue #135, opened 2012-03-14
Problem with second layer of a 2 layer coded LRCP (with precincts)


gcode-importer:
Originally reported on Google Code with ID 135
```
What steps will reproduce the problem?
1. j2k_to_image -l 2 -i test-12.j2c -o test-12.tif
2. (see attached test-12.j2c)
3.

What is the expected output? What do you see instead?
The program does not complain, but the output is
wrongly decoded, with a lot of garbage. 

What version of the product are you using? On what operating system?
openjpeg 1.5.0 on fedora 16 (3.2.9-2.fc16.x86_64)

Please provide any additional information below.

Decoding only the first layer does work correctly. 
Decondig with kakadu works correctly for both layers.
I think the codestream is legal, but I am not 100% sure.

```

Reported by `luca.trisciani` on 2012-03-14 16:07:08

<hr>
 * *Attachment: [test-12.j2c](https://storage.googleapis.com/google-code-attachments/openjpeg/issue-135/comment-0/test-12.j2c)*

szukw000:
I have found that RaspberryPi ( a 32-Bit machine ) does show the KODAK
image of issue135 100% fine with both layers. 'opj_decompress' on the
RaspberryPi decompresses all images 100% fine (except TGA: too dark).

LINUX 32-Bit does show the KODAK image broken. 'opj_decompress' on
LINUX 32-Bit decompresses all images broken.

winfried

mayeut:
By changing line 1408 of t1.c from
```c
  for (passno = 0; passno < seg->real_num_passes; ++passno) {
```
to
```c
  for (passno = 0; (passno < seg->real_num_passes) && (bpno_plus_one >= 1); ++passno) {
```
File seem to decode properly (as well as test_lossless.j2k from the test suite).

I tried this following report from UBSan on CDash: http://my.cdash.org/viewDynamicAnalysisFile.php?id=3209124
I don't know if this is a legitimate correction but thought I'd share what I found.

szukw000:
@mayeut ,

the change in line 1408 of t1.c was successful:
on LINUX 64-Bit
on Win7 64-Bit
on PPC64

RaspberryPi I could not test because of:

CMake Error at src/lib/openjp2/CMakeLists.txt:88 (target_compile_options):
  Unknown CMake command "target_compile_options".


-- Configuring incomplete, errors occurred!

winfried

mayeut:
@szukw000,

[...]

PR #706, opened 2016-01-25
Fix issue 135

mayeut:
@detonin,

This is the PR you asked for in #135 
As I said in the issue, I don't know if this is something legal regarding the standard (or if the file is broken somehow)


detonin:
The fix is legal regarding the standard but I did not manage to find out if it covers a bug in opj_t2_read_packet_data or if the file is corrupted

PR #708, opened 2016-01-29
issue #695 MQ Encode: ensure that bp pointer never points to uninitialized memory

### LINKED ISSUE #695: MQ Coder encode: Conditional jump or move depends on uninitialised value(s)

boxerab:
fixes #695

### DIFF
--- a/src/lib/openjp2/mqc.c
+++ b/src/lib/openjp2/mqc.c
@@ -203,13 +203,14 @@ static opj_mqc_state_t mqc_states[47 * 2] = {
 */
 
 static void opj_mqc_byteout(opj_mqc_t *mqc) {
-	if (*mqc->bp == 0xff) {
+	OPJ_BYTE bp_in_bounds = (mqc->bp >= mqc->start);
+	if (bp_in_bounds & (*mqc->bp == 0xff)) {
 		mqc->bp++;
 		*mqc->bp = (OPJ_BYTE)(mqc->c >> 20);
 		mqc->c &= 0xfffff;
 		mqc->ct = 7;
 	} else {
-		if ((mqc->c & 0x8000000) == 0) {	/* ((mqc->c&0x8000000)==0) CHANGE */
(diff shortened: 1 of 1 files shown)


boxerab:
This has to fix the uninitialized memory issue. What I find surprising is that such a critical bug has existed for so long ( > 5 years )

mayeut:
@boxerab, Unfortunately, this doesn't help: http://my.cdash.org/viewDynamicAnalysisFile.php?id=3438342

boxerab:
The patch fixes a definite bug: there may be other similar bugs in the MQ coder that continue causing a valgrind error.

detonin:
@boxerab @mayeut does this fix #612 ?

julienmalik:
I'm currently looking into refreshing the dashboard submissions from CS, which are currently only noise since the GitHub migration, because they are still looking for updates from svn :)

Seeing all the work achieve with Travis, I will remove most of them, but try to setup again the valgrind build where we should see the effect of this patch.

boxerab:
Unfortunately, I don't have time to look into this at the moment.

julienmalik:
@detonin : no it does not fix #612 : http://my.cdash.org/viewDynamicAnalysisFile.php?id=3567436

PR #714, opened 2016-02-17
Only convert to RGB colorspace when requested

slugan:


mayeut:
@slugan 
In order to keep a backward compatible behavior, could you add the inverse option instead. Something like `--no-rgb-conversion`

slugan:
@mayeut
The backward-compatible behavior is to use the current `-force-rgb` flag (introduced in https://github.com/uclouvain/openjpeg/commit/7e3844607584141d438191433c700d79ffab3b2a ) when a non-RGB color space should be converted to RGB (no color space conversion is supposed to be performed by the codec by default).

The proposed changeset fixes a bug introduced in https://github.com/uclouvain/openjpeg/commit/238f4419abb035200ce1cf56f9da677f5eca8afb where extra color space conversions to RGB have been inserted out of the `if (parameters.force_rgb)` test.

Introducing a new `--no-rgb-conversion` would be contradictory with the currently implemented `-force-rgb` flag (which should then be removed) and would reverse OpenJPEG's current default behavior.

mayeut:
@slugan sorry I missed that the fix was only for those colorspaces.

Anyway, @detonin shall say if consistent behavior shall be kept for those colorspaces (i.e. convert to RGB as default like other colorspaces except grayscale which requires `--force-rgb`) & thus consider a bug existed before 238f441 for those newly handled colorspaces. Or if backward-compatible behavior shall be enforced as proposed in this PR (in which case the tests will need to be updated in order to pass ci checks).

detonin:
@mayeut I do not see any reason why we would actually convert by default from a given colorspace (provided in a JP2 box) to RGB. If a specific colorspace is set in the JP2 file, we should keep it and, if possible, signal it in the output format (and on stdout maybe). 

So the following tests would need to be updated:
ETS-RIC-subsampling_1.jp2-compare2ref
ETS-RIC-zoo1.jp2-compare2ref
NR-DEC-merged.jp2-19-decode-md5
NR-DEC-issue142.j2k-66-decode-md5
NR-DEC-issue411-ycc444.jp2-113-decode-md5
NR-DEC-issue411-ycc422.jp2-114-decode-md5
NR-DEC-issue411-ycc420.jp2-115-decode-md5

PR #729, opened 2016-03-24
[fix] Allow to convert 3/5/7/9/11/13/15 bpp JP2 files to TIF

julienmalik:
This issue comes from the decompression of Sentinel-2 Earth Observation satellite products, which are currently delivered as 15-bpp JP2 files.

OpenJPEG 2.0.1 and 2.1 are affected by a serious bug when using opj_decompress to convert the JP2 files to TIF format : the actual pixels values are twice the expected values.

Current trunk simply refuses to convert the data to TIF and errors out with the message :
```
imagetotif: Bits=0, Only 1, 2, 4, 6, 8, 10, 12, 14 and 16 bits implemented
```

This commit simply considers the unsupported 3/5/7/9/11/13/15 bit-precision as if they were 4/6/8/10/12/14/16 bpp data, to at least allow TIF conversion. The bug affecting 2.0.1 and 2.1 cannot be reproduced then, and pixel values in TIF are as expected.

Please note that Kakadu does output a 15 bpp TIF :
```
$ tiffinfo S2A_OPER_MSI_L1C_TL_SGS__20151127T035842_A002245_T52SEG_B08_KDU.tif
TIFF Directory at offset 0x8 (8)
  Image Width: 10980 Image Length: 10980
  Resolution: 0.01, 0.01 (unitless)
  Bits/Sample: 15
  Sample Format: unsigned integer
  Compression Scheme: None
  Photometric Interpretation: min-is-black
  Samples/Pixel: 1
  Rows/Strip: 814
  Planar Configuration: single image plane
```

while with this fix, OpenJPEG outputs a 16bpp TIF :
```
tiffinfo S2A_OPER_MSI_L1C_TL_SGS__20151127T035842_A002245_T52SEG_B08_OPJ_TRUNK.tif
TIFF Directory at offset 0xe5f3628 (241120808)
  Image Width: 10980 Image Length: 10980
  Bits/Sample: 16
  Compression Scheme: None
  Photometric Interpretation: min-is-black
  Orientation: row 0 top, col 0 lhs
  Samples/Pixel: 1
  Rows/Strip: 1
  Planar Configuration: single image plane
```

I do have some issues with the 15bpp TIFF files in other software (namely the Java JAI TIF reader), so for my use case it is better to keep only 1, 2, 4, 6, 8, 10, 12, 14 and 16 bits TIF.

### DIFF
--- a/src/bin/jp2/converttif.c
+++ b/src/bin/jp2/converttif.c
@@ -166,7 +166,7 @@ static void tif_32sto16u(const OPJ_INT32* pSrc, OPJ_UINT16* pDst, OPJ_SIZE_T len
 int imagetotif(opj_image_t * image, const char *outfile)
 {
 	int width, height;
-	int bps,adjust, sgnd;
+	int bps,adjust, sgnd, tifbps;
 	int tiPhoto;
 	TIFF *tif;
 	tdata_t buf;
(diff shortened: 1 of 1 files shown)


boxerab:
Cool : I saw this problem with sentinel images. 15 bpp was failing.

boxerab:
It might be wise to add an odd bit depth JP2 image to the test suite.

detonin:
@julienmalik I agree with Aaron. Could you provide such odd bit depth JP2 image so that we can add it to the test suite and reproduce the bug before fixing it ? Best way to do so is to create a PR in openjpeg-data with the new file and then add the test in https://github.com/uclouvain/openjpeg/blob/master/tests/nonregression/test_suite.ctest.in
thanks

julienmalik:
Agreed

julienmalik:
After further testing, seems like 12bpp TIFF is not supported by all implementation, so it could help to have a command line option to force the bitdepth of the output. I'll try to come up with something.

mayeut:
@julienmalik, the command line option to force output bit depth already exists.

```
-p <comp 0 precision>[C|S][,<comp 1 precision>[C|S][,...]]
    OPTIONAL
    Force the precision (bit depth) of components.
    There shall be at least 1 value. Theres no limit on the number of values (comma separated, last values ignored if too much values).
    If there are less values than components, the last value is used for remaining components.
    If 'C' is specified (default), values are clipped.\n"
    If 'S' is specified, values are scaled.\n"
    A 0 value can be specified (meaning original bit depth).\n
```

julienmalik:
many thanks @mayeut ! much more than actually expected !

PR #739, opened 2016-04-20
cppcheck fix for openmj2

julienmalik:


rouault:
Obsolete since that component has now been removed

PR #740, opened 2016-04-20
cppcheck fix for openjp2

julienmalik:


src/bin/common/color.c
```diff
@@ -116,6 +118,12 @@ static void sycc444_to_rgb(opj_image_t *img)
 	free(img->comps[0].data); img->comps[0].data = d0;
 	free(img->comps[1].data); img->comps[1].data = d1;
 	free(img->comps[2].data); img->comps[2].data = d2;
+	return;
+
+fails:
+	if(r) free(r);
```
julienmalik: free can be called on a NULL pointer (see section 7.20.3.2 of http://www.open-std.org/JTC1/SC22/wg14/www/docs/n1124.pdf)

PR #752, opened 2016-04-25
Fix implicit floating bool conversion

julienmalik:
clang reports a number of implicit float-to-bool conversions.

PR #753, opened 2016-04-25
Add clang 3.9 build to Travis matrix

julienmalik:


julienmalik:
Updates #752

julienmalik:
With the latest fix, the build now runs OK in travis with clang-3.9.

Only some tests are failing :
https://travis-ci.org/uclouvain/openjpeg/jobs/125578218#L3794-L3801
while they don't fail with gcc or clang-3.4

julienmalik:
MemorySanitizer has improved in clang between between 3.4 and 3.9. My final aim was to set up a msan build in travis to track down introduction of bad code (in addition to catching a bit more warnings). With clang-3.9 it is possible to get the stack trace easily.

Also I found it has some advantage over valgrind : quicker in general (3x, and additionnal 2x with symbolizing the output), and execution of tests can be stopped as soon as a defect is detected.

mayeut:
I've got the same tests failing on my setup. Can you add the correct exception list for clang-3.9 so that the PR shows successful under Travis CI ?

julienmalik:
@mayeut Now it passes with the known failures exceptions added. I rebased and squashed everything.

Not sure though if keeping both release and debug builds is relevant/useful.

mayeut:
@julienmalik, Thanks.

The matrix is already quite big and, as you said, keeping both release & debug configurations might not be relevant/useful.

I think you could remove either one of them (preferably remove the debug build to get faster tests) or replace the existing ASan build with clang-3.9

PR #755, opened 2016-04-25
Add the build number to CTEST_BUILD_NAME to differenciate builds on d…

julienmalik:
…ashboard


julienmalik:
This allows to find the travis/appveyor build from the dashboard build name

julienmalik:
Later I would like to add a notes file to the dashboard with direct links to travis build page.

julienmalik:
ping @mayeut @detonin

julienmalik:
Sample build with travis build number inside its name :  http://my.cdash.org/buildSummary.php?buildid=949172

mayeut:
Even though it seems convenient, it prevents the dashboard from updating the passed/failed tests difference for a given PR. The notes file seems like a better idea to me.

detonin:
Would be nice indeed to keep track of passed/failed tests updates. 
I see that the build note contains the whole travis-ci.cmake file (http://my.cdash.org/viewNotes.php?buildid=949172). Maybe is there a way to append the build number there.
Concerning the bot post, I think it's a good idea as the dashboard gives additional inputs compared to the github checks status board. 
Many thanks.

julienmalik:
For the bot post, it is not easy to implement since it would require the bot credentials for posting a comment. These need to be encrypted in the travis.yml to not expose the GitHub credentials of the bot, but encrypted env vars are not available for pull requests, for security reasons.

Issue #718, opened 2016-03-02
fuzzing-created crashers


setharnold:
Hello; I reviewed OpenJPEG recently for the Ubuntu Linux distribution. As part of this effort I ran our currently packaged 1.5.2 version through the AFL fuzzer for roughly a day and found several issues. I verified that these specific crashing files do not crash the 2.1.0 upstream release, but four hours of fuzzing time found a crashing file on 2.1.0.

I've uploaded the tarball of files that caused crashes to Launchpad:
https://bugs.launchpad.net/ubuntu/+source/openjpeg/+bug/711061/+attachment/4586223/+files/openjpeg-crashers.tar.gz
If that URL doesn't work, the bug is at:
https://bugs.launchpad.net/ubuntu/+source/openjpeg/+bug/711061

Please consider adding these files to your test suite.

When you've had an opportunity to diagnose and prepare fixes, it would be hugely beneficial to downstream consumers if you could also request CVEs on the oss-security mail list:
http://oss-security.openwall.org/wiki/mailing-lists/oss-security

I would be happy to help request CVEs when you're ready.

Thanks

szukw000:
A first test:

szukw000: gm animate relax.jp2     
ICC Profile CS 52474220
gm: jp2_dec.c:302: jp2_decode: Assertion `dec->image->cmprof_' failed.
gm animate: abort due to signal 6 (SIGABRT) "Abort"...
Aborted

szukw000: animate relax.jp2     

szukw000: flimage relax.jp2     

A second test:


diff -u relax.jp2 id\:000000\,sig\:11\,src\:000008\,op\:flip2\,pos\:354.jp2 > test.dif

//vim test.dif:
Binary files relax.jp2 and id:000000,sig:11,src:000008,op:flip2,pos:354.jp2 differ

A second test:

szukw000: gm animate id\:000000\,sig\:11\,src\:000008\,op\:flip2\,pos\:354.jp2
error: no code stream found
gm animate: Unable to decode image file (id:000000,sig:11,src:000008,op:flip2,pos:354.jp2).

szukw000: animate id\:000000\,sig\:11\,src\:000008\,op\:flip2\,pos\:354.jp2 
animate: Box length is inconsistent.
 `OpenJP2' @ error/jp2.c/JP2ErrorHandler/193.
animate: Stream error while reading JP2 Header box
 `OpenJP2' @ error/jp2.c/JP2ErrorHandler/193.
[...]

PR #761, opened 2016-04-28
Fix unsigned int overflow reported by UBSan

mayeut:
Please add -DOPJ_UBSAN_BUILD to CFLAGS when building with
-fsanitize=undefined,unsigned-integer-overflow

It seems clang/gcc do not allow to disable checking for block of code
other than function or file.

Update #718

### DIFF
--- a/src/lib/openjp2/bio.c
+++ b/src/lib/openjp2/bio.c
@@ -151,19 +151,31 @@ void opj_bio_init_dec(opj_bio_t *bio, OPJ_BYTE *bp, OPJ_UINT32 len) {
 	bio->ct = 0;
 }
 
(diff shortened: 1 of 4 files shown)


julienmalik:
I missed this.

Now the nightly dashboard submission from dora.c-s.fr for ubsan will include this -DOPJ_UBSAN_BUILD (both gcc and clang).
This removes 2 ubsan reports for the clang build.

Though I'm not sure to understand why we end up in the assert while the file should be rejected (should openpeg fail earlier ?).


Also I don't understand the your comment "It seems clang/gcc do not allow to disable checking for block of code other than function or file." : opj_bio_read/write get inlined so the OPJ_NOSANITIZE() does not work ?

PR #765, opened 2016-05-01
Use git submodule for third party libraries

mayeut:
This will help with future updates of liblcms2


julienmalik:
CMake external project is another alternative, and I just thought about it this afternoon...

Already used it successfully in some other projects.
It is good because it keeps the cmake cache of all projects independent.

For tiff and lcms, which are cmake projects, this is for sure a good alternative. might not be a little less easy to set up for png (to keep an easy build on windows)

mayeut:
I haven't used CMake external project yet. That might be another (perhaps better) option.
Nevertheless, we should rely on official sources from those libs & it's easier to update & keep track of using submodules.
PNG & ZLIB are also provide cmake builds.

mayeut:
This will not work as is with minimum cmake version. Closing & reopening a new one later.

PR #767, opened 2016-05-02
Fix UBSan gcc warning for first arg to memset non null

julienmalik:


julienmalik:
I manually checked that the NULL case was when w=1 and h=0, so we were actually memsetting zero bytes.

Maybe this should be detected earlier ?

julienmalik:
Updates #718

julienmalik:
Only 5 UB left on gcc 5.2 : http://my.cdash.org/viewDynamicAnalysis.php?buildid=953384

PR #768, opened 2016-05-02
Ignore clang's summary warning

julienmalik:
This assumes prior text has matched some other warning expression.

This reduces the warnings reported for clang build on the dashboard,
which were caused only by the "XXX warnings generated." message.
For some reason they were not reported when not using ctest launchers.

This commit allows to confidently use ctest launchers to improve
dashboard reports.

### DIFF
--- a/cmake/CTestCustom.cmake.in
+++ b/cmake/CTestCustom.cmake.in
@@ -25,6 +25,9 @@ set(CTEST_CUSTOM_COVERAGE_EXCLUDE
 set(CTEST_CUSTOM_WARNING_EXCEPTION
   ${CTEST_CUSTOM_WARNING_EXCEPTION}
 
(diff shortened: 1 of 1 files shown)


julienmalik:
Now I get, with clang 3.9 on linux, a single warning report :
http://my.cdash.org/viewBuildError.php?type=1&buildid=953326

which is most probably coming from :

```
/usr/include/assert.h:92:5: note: expanded from macro 'assert'
  ((expr)                                                               \
```

We could include this as an exception too, but I'm not sure if it is the best move.
Maybe disabling completely the warnings at compile time for thirdparties ? At least when a specific cmake advanced option is ON (OPJ_THIRDPARTY_NO_WARNING or something like this), which we can use on the dashboard ?

PR #771, opened 2016-05-02
Use git submodules for zlib, libpng & liblcms2

mayeut:
This will ease updates of third party libraries.
An unofficial mirror of libtiff exists here: https://github.com/vadz/libtiff
Not sure if it should be included. Leaving old copy for now.

It might also be a good thing to build as external projects with cmake.


boxerab:
There are a lot of disadvantages to using git submodules, just saying:
http://somethingsinistral.net/blog/git-submodules-are-probably-not-the-answer/

For third party libraries, it probably makes sense, but they can be a pain or downright dangerous to use.

Personally, I would avoid them - third party libraries should probably only be updated when they get released, which doesn't happen very often.

mayeut:
@boxerab,

I quote the article:

>**What submodules are for**
> ...
> Three - There’s a part of your repository that you’re delegating to another party. Let’s say you’re paying someone to write a plugin for a project you’re using, and you need to develop on the main codebase. In this case, the plugin repository is chiefly developed by the plugin developers, so they own the repo and periodically they’ll tell you when to update submodule commits. Submodules are great for dividing responsibilities like this, assuming that there’s not frequent updating.

So, and as you said, it does make sense for third party libraries.
That being said, I don't think they'll be dangerous in this case. However, git submodules are not the easiest to work with & a few commands are needed to begin with.

I'm open to other suggestions. @julienmalik talked about external project support in cmake, that might be a good solution.

As of now, those third party libraries are embedded in the repo & eventually modified (which is bad - for those specific libraries and their use in this specific project - and troublesome to update).

I'm not quite happy with what I'm proposing either because it still needs a specific `CMakeLists.txt` but, nevertheless, I think it's better than what's existing. Again, maybe external project is a better solution but I'm not familiar with it.

stweil:
I made good experiences with submodules in [QEMU](https://github.com/qemu/qemu). Like OpenJPEG, that project uses submodules for external projects which are not needed for everybody. The big advantage is that the submodule code is only pulled if needed, that it is clearly defined where it comes from, that it is easy to update a submodule (not only for the official repository but also for local tests), that the full history of a submodule is visible and maybe more.

@mayeut, may I suggest to use the old path names like `thirdparty/libpng` for the submodules instead of the lengthy new ones like `thirdparty/libpng/libpng`?

mayeut:
@stweil, I used the old path names in another closed PR. The issue is that, given that OpenJpeg uses specific `CMakeLists.txt`for each third party libraries, all of those get merged at `thirdparty/CMakeLists.txt` level. I used specific cmake commands so that there were as little interaction between the libraries since there's only one cmake scope in that case. Those commands are not available to cmake 2.8.x and that's why I reverted to this layout.
Now that I think about it, it might be possible to use the `function` command that creates a new scope.
I will try this.

mayeut:
@stweil, I updated zlib only for now with a simpler layout. If that seems ok to you, I will apply it to the others as well.

mayeut:
@stweil, after applying the trick to libpng, seems that `include_directories` doesn't get scoped. I will revert to the original layout. I tried working that around but doesn't work with cmake 2.8.7.

stweil:
I am not am expert for cmake, so I don't know how to realize a solution with less directory levels.

kbabioch:
There is an [official](https://gitlab.com/libtiff/libtiff) libtiff Git now.

PR #772, opened 2016-05-03
Allow using CTEST_USE_LAUNCHERS even when tests are disabled

julienmalik:
Since cmake >= 3.0, using CTEST_USE_LAUNCHERS requires include(CTestUseLaunchers),
which is automatically done by include(CTest)

Setting CTEST_USE_LAUNCHERS in the cache without including CTest later
triggers a configuration error :
   CMake Error: CTEST_USE_LAUNCHERS is enabled, but the RULE_LAUNCH_COMPILE
   global property is not defined.
   Did you forget to include(CTest) in the toplevel CMakeLists.txt ?

CTEST_USE_LAUNCHERS was silently ignored in cmake <= 2.8 when not including CTest

Include CTestUseLaunchers always, so that CTEST_USE_LAUNCHERS can be used even when not
including CTest (for example for MinGW builds).
This helps presenting the build errors and warnings more cleanly on the Dashboard.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -318,6 +318,9 @@ if(BUILD_TESTING)
   endif()
 endif()
 
+# Allow the use of CTEST_USE_LAUNCHERS even when not including CTest (mingw builds)
+include(CTestUseLaunchers OPTIONAL)
+
 #-----------------------------------------------------------------------------
 # install all targets referenced as OPENJPEGTargets
 install(EXPORT OpenJPEGTargets DESTINATION ${OPENJPEG_INSTALL_PACKAGE_DIR})


julienmalik:
Hum... quite strange that the mingw warnings disappear with ctest launchers enabled. This should be investigated further.

julienmalik:
CTEST_USE_LAUNCHERS  is even ignored for this mingw build since it uses ctest 2.8, which ignores CTEST_USE_LAUNCHERS since include(CTest) is not used. We need cmake >= 3.0 here

julienmalik:
With cmake 3.5 it gets better : http://my.cdash.org/viewBuildError.php?type=1&buildid=953977
The warnings are caught and the dashboard report is clear.

julienmalik:
@mayeut ok for merging ?

might want to still have a cmake 2.8 build around though (I see that appveyor also uses cmake 3.5)...

mayeut:
@julienmalik,

If I understand correctly, there's no more build using cmake 2.8.x.

`find . -name 'CMakeLists.txt' -exec grep 'cmake_minimum_required' {} +` outputs the following:
```
./CMakeLists.txt:cmake_minimum_required(VERSION 2.8.2)
./tests/nonregression/CMakeLists.txt:cmake_minimum_required(VERSION 2.8.7)
./thirdparty/libtiff/CMakeLists.txt:cmake_minimum_required(VERSION 2.6)
./thirdparty/libz/CMakeLists.txt:cmake_minimum_required(VERSION 2.6)
```
It would be good to keep at least one building with cmake 2.8.x. (e.g. I used a feature from cmake 3.x in a PR and didn't realize it until CI checks failed).
I run tests using cmake 2.8.7 once in a while but not frequently enough to assess anything.

PR #790, opened 2016-06-06
Update tcd.c

maddin200:
cppcheck uninitialized variable


maddin200:
Thanks for considering.

PR #818, opened 2016-09-05
Fix leak & invalid behavior of opj_jp2_read_ihdr

mayeut:
In case multiple ihdr box are present, only the first one shall be
taken into account.


mayeut:
https://pdfium.googlesource.com/pdfium/+/225a257ba299a43c02471fa46c22ba94da1f3f09

PR #823, opened 2016-09-07
Add sanity check for tile coordinates

mayeut:
Coordinates are casted from OPJ_UINT32 to OPJ_INT32
Add sanity check for negative values and upper bound becoming lower
than lower bound.
See also
https://pdfium.googlesource.com/pdfium/+/b6befb2ed2485a3805cddea86dc7574510178ea9


mayeut:
there was an error when referencing test data.
Named issue822 #822 when it should be issue823

nluedtke:
http://www.openwall.com/lists/oss-security/2016/09/08/8 indicates this issue was the same as CVE-2016-5157

PR #824, opened 2016-09-07
Fix coverity 113065 (CWE-484)

mayeut:
@detonin,

I'm almost sure the fall through is intentional otherwise nothing would work but, just to be sure, can you confirm ?
The added comment will tell coverity not to report an issue & the code will be clearer for everyone.

### DIFF
--- a/src/lib/openjp2/t1_generate_luts.c
+++ b/src/lib/openjp2/t1_generate_luts.c
@@ -52,6 +52,7 @@ static int t1_init_ctxno_zc(int f, int orient) {
 			t = h;
 			h = v;
 			v = t;
+			/* fall through */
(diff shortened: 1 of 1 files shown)


detonin:
@mayeut yes this is intentional and can be merged.
What I actually wonder is if we need to keep t1_generate_luts where it currently is. It has been used to generate a static file (t1_luts.h) but there is no need to compile it each time we compile OpenJPEG: I think we should keep t1_luts.h and move t1_generate_luts in the tools folder for instance.

mayeut:
@detonin, agreed for the file location. I'll let you create an issue for this as a reminder.

detonin:
@mayeut #831

PR #825, opened 2016-09-08
Strengthen integer overflow check in opj_pi_create_decode

trylab:
The crash still happens!!!

``l_tcp->numlayers`` and ``l_step_l`` are both **OPJ_UINT32** type variables. Thus using ``SIZE_MAX`` or ``((size_t)-1)`` to check integer overflow is insufficient. We should use ``(OPJ_UINT32)-1`` here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,17 @@
+# Ignore files and directories made by `cmake .`.
+CMakeFiles/
+Makefile
+cmake_install.cmake
+/CMakeCache.txt
+/CPackConfig.cmake
+/CPackSourceConfig.cmake
+/CTestCustom.cmake
+/LICENSE.txt
+/OpenJPEGConfig.cmake
+/libopenjp2.pc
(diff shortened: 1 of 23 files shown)


stweil:
Could you please test the current code with this modification:

    l_current_pi->include = (OPJ_INT16*) opj_calloc(((size_t)l_tcp->numlayers +1) * l_step_l, sizeof(OPJ_INT16)); 

The type cast is needed because otherwise addition and multiplication are done using 32 bit. The result is converted to size_t and passed to opj_calloc.

mayeut:
@stweil is right
See ef01f18dfc6780b776d0674ed3e7415c6ef54d24

trylab:
@mayeut I opened (and closed) a issue to track it. You can get a POC at https://github.com/uclouvain/openjpeg/issues/826

mayeut:
Adding data to the test suite in progress.
Thanks

Issue #780, opened 2016-05-06
division-by-zero in function opj_pi_next_rpcl of pi.c (line 366)


trylab:
# Title

division-by-zero in function opj_pi_next_rpcl of pi.c (line 366)
# Testing Environment

Ubuntu + OpenJPEG (GitHub master, 2016/05/06)
# Exception Information

```
Program received signal SIGFPE, Arithmetic exception.
0xb69892b1 in opj_pi_next_rpcl (pi=0xb3a03ec0) at 
    /home/trylab/Desktop/repo/openjpeg/src/lib/openjp2/pi.c:366
366 if (!((pi->x % (OPJ_INT32)(comp->dx << rpx) == 0) || 
        ((pi->x == pi->tx0) && ((trx0 << levelno) % (1 << rpx))))){
```
# PoC

https://raw.githubusercontent.com/trylab/PoCs/master/openjpeg/SIGFPE_opj_pi_next_rpcl@366/poc.j2k
# Credit

Ke Liu of Tencent's Xuanwu LAB

szukw000:
bin/opj_decompress  -i /tmp/ISSUE-780/issue780-poc.j2k -o issue780-poc.j2k.png

\[INFO\] Start to read j2k main header (0).
\[INFO\] Main header has been correctly decoded.
\[INFO\] No decoded area parameters, set the decoded area to the whole image
\[INFO\] Stream reached its end !
\[INFO\] Header of tile 6 / 64 has been read.
Floating point exception

NAME(/tmp/ISSUE-780/issue780-poc.j2k)
LENG(568)

ENTER read_jp2c
\[0\]marker(0xff4f)
    soc len(0)
\[2\]marker(0xff51)
    siz len(47)
    capabilities(0)\[extended: 0\]
    x(0 : 64) y(0 : 64)
    xt(0 : 64) yt(0 : 1)
    IMAGE w(64) h(64) TILE w(64) h(1)

    ----------------------

Tested with openjpeg-master-2016-05-09.

winfried

rouault:
Fixed per https://github.com/uclouvain/openjpeg/commit/d27ccf01c68a31ad62b33d2dc1ba2bb1eeaafe7b

Issue #779, opened 2016-05-06
division-by-zero in function opj_pi_next_rpcl of pi.c (line 363)


trylab:
# Title

division-by-zero in function opj_pi_next_rpcl of pi.c (line 363)
# Testing Environment

Ubuntu + OpenJPEG (GitHub master, 2016/05/06)
# Exception Information

```
Program received signal SIGFPE, Arithmetic exception.
0xb69891a6 in opj_pi_next_rpcl (pi=0xb4603d80) 
    at /home/trylab/Desktop/repo/openjpeg/src/lib/openjp2/pi.c:363
363 if (!((pi->y % (OPJ_INT32)(comp->dy << rpy) == 0) || 
        ((pi->y == pi->ty0) && ((try0 << levelno) % (1 << rpy))))){
```
# PoC

https://raw.githubusercontent.com/trylab/PoCs/master/openjpeg/SIGFPE_opj_pi_next_rpcl@363/poc.j2k
# Credit

Ke Liu of Tencent's Xuanwu LAB

rouault:
Fixed per https://github.com/uclouvain/openjpeg/commit/d27ccf01c68a31ad62b33d2dc1ba2bb1eeaafe7b

Issue #732, opened 2016-03-28
division-by-zero (SIGFPE) error in opj_pi_next_cprl function (line 526 of pi.c)


trylab:
---
## Testing Environment

Ubuntu + OpenJPEG (GitHub master, 2016/03/28)

---
## Exception Information

username@ubuntu:~/Desktop/openjpeg/bin$ gdb opj_decompress -q
Reading symbols from opj_decompress...done.
(gdb) r -o image.pgm -i crashes/002.j2k 
Starting program: ~/Desktop/openjpeg/bin/opj_decompress -o image.pgm -i crashes/002.j2k

Program received signal SIGFPE, Arithmetic exception.
0xb7fb8d01 in opj_pi_next_cprl (pi=0x8090ec0) 
    at /home/username/Desktop/openjpeg/src/lib/openjp2/pi.c:526
526  if (!((pi->x % (OPJ_INT32)(comp->dx << rpx) == 0) || 
        ((pi->x == pi->tx0) && ((trx0 << levelno) % (1 << rpx))))){

(gdb) p comp->dx
$1 = 2
(gdb) p rpx
$2 = 31

(gdb) bt
#0  0xb7fb8d01 in opj_pi_next_cprl (pi=0x8090ec0)

```
at /home/username/Desktop/openjpeg/src/lib/openjp2/pi.c:526
```
#1  0xb7fbc4cc in opj_pi_next (pi=0x8090ec0)

```
at /home/username/Desktop/openjpeg/src/lib/openjp2/pi.c:1871
```
#2  0xb7fc0b8d in opj_t2_decode_packets (p_t2=0x8090878, p_tile_no=0, p_tile=0x80658d0,

```
                                     p_src=0x80668d8 "\337\aV", p_data_read=0xbfff9dec, 
                                     p_max_len=49, p_cstr_index=0x8061d68, p_manager=0x80601e4)
at /home/username/Desktop/openjpeg/src/lib/openjp2/t2.c:412
```
#3  0xb7fc6368 in opj_tcd_t2_decode (p_tcd=0x8065890, p_src_data=0x80668d8 "\337\aV",

```
                                 p_data_read=0xbfff9dec, p_max_src_size=225, 
                                 p_cstr_index=0x8061d68, p_manager=0x80601e4) 
at /home/username/Desktop/openjpeg/src/lib/openjp2/tcd.c:1552
```
[...]

Issue #731, opened 2016-03-28
division-by-zero (SIGFPE) error in opj_pi_next_cprl function (line 523 of pi.c)


trylab:
---
## Testing Environment

Ubuntu + OpenJPEG (GitHub master, 2016/03/28)

---
## Exception Information

```
Program received signal SIGFPE, Arithmetic exception.
0xb7fb8ca1 in opj_pi_next_cprl (pi=0x80850d8) 
    at /home/username/Desktop/openjpeg/src/lib/openjp2/pi.c:523
523  if (!((pi->y % (OPJ_INT32)(comp->dy << rpy) == 0) || 
        ((pi->y == pi->ty0) && ((try0 << levelno) % (1 << rpy))))){

(gdb) p rpy
$4 = 26
(gdb) p comp->dy
$5 = 128

(gdb) bt
#0  0xb7fb8ca1 in opj_pi_next_cprl (pi=0x80850d8) 
    at /home/username/Desktop/openjpeg/src/lib/openjp2/pi.c:523
#1  0xb7fbc4cc in opj_pi_next (pi=0x80850d8) 
    at /home/username/Desktop/openjpeg/src/lib/openjp2/pi.c:1871
#2  0xb7fc0b8d in opj_t2_decode_packets (p_t2=0x8084a90, p_tile_no=0, 
                                         p_tile=0x80658d0, p_src=0x80668d8 "\337\aV", 
                                         p_data_read=0xbfff9dec, p_max_len=0, 
                                         p_cstr_index=0x8061d68, p_manager=0x80601e4) 
    at /home/username/Desktop/openjpeg/src/lib/openjp2/t2.c:412
#3  0xb7fc6368 in opj_tcd_t2_decode (p_tcd=0x8065890, p_src_data=0x80668d8 "\337\aV", 
                                     p_data_read=0xbfff9dec, p_max_src_size=225, 
                                     p_cstr_index=0x8061d68, p_manager=0x80601e4) 
    at /home/username/Desktop/openjpeg/src/lib/openjp2/tcd.c:1552
#4  0xb7fc5d17 in opj_tcd_decode_tile (p_tcd=0x8065890, p_src=0x80668d8 "\337\aV", 
                                       p_max_length=225, p_tile_no=0, 
                                       p_cstr_index=0x8061d68, p_manager=0x80601e4) 
    at /home/username/Desktop/openjpeg/src/lib/openjp2/tcd.c:1291
[...]

PR #845, opened 2016-09-21
Fix some division-by-zero problems in src\lib\openjp2\pi.c

trylab:
The following issues were fixed in this commit.

https://github.com/uclouvain/openjpeg/issues/731
https://github.com/uclouvain/openjpeg/issues/732
https://github.com/uclouvain/openjpeg/issues/777
https://github.com/uclouvain/openjpeg/issues/778
https://github.com/uclouvain/openjpeg/issues/779
https://github.com/uclouvain/openjpeg/issues/780

### DIFF
--- a/src/lib/openjp2/pi.c
+++ b/src/lib/openjp2/pi.c
@@ -360,8 +360,14 @@ if (!pi->tp_on){
 					try1 = opj_int_ceildiv(pi->ty1, (OPJ_INT32)(comp->dy << levelno));
 					rpx = res->pdx + levelno;
 					rpy = res->pdy + levelno;
+					if ((OPJ_INT32)(comp->dy << rpy) == 0 || (1 << rpy) == 0) {
+						continue;
+					}
 					if (!((pi->y % (OPJ_INT32)(comp->dy << rpy) == 0) || ((pi->y == pi->ty0) && ((try0 << levelno) % (1 << rpy))))){
 						continue;	
+					}
+					if ((OPJ_INT32)(comp->dx << rpx) == 0 || (1 << rpx) == 0) {
+						continue;
 					}
 					if (!((pi->x % (OPJ_INT32)(comp->dx << rpx) == 0) || ((pi->x == pi->tx0) && ((trx0 << levelno) % (1 << rpx))))){
 						continue;
@@ -441,8 +447,14 @@ static OPJ_BOOL opj_pi_next_pcrl(opj_pi_iterator_t * pi) {
 					try1 = opj_int_ceildiv(pi->ty1, (OPJ_INT32)(comp->dy << levelno));
 					rpx = res->pdx + levelno;
 					rpy = res->pdy + levelno;
+					if ((OPJ_INT32)(comp->dy << rpy) == 0 || (1 << rpy) == 0) {
+						continue;
+					}
 					if (!((pi->y % (OPJ_INT32)(comp->dy << rpy) == 0) || ((pi->y == pi->ty0) && ((try0 << levelno) % (1 << rpy))))){
 						continue;	
+					}
+					if ((OPJ_INT32)(comp->dx << rpx) == 0 || (1 << rpx) == 0) {
+						continue;
 					}
 					if (!((pi->x % (OPJ_INT32)(comp->dx << rpx) == 0) || ((pi->x == pi->tx0) && ((trx0 << levelno) % (1 << rpx))))){
 						continue;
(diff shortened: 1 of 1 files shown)


mayeut:
This would also fix some issues with images from #811 

@detonin, Is the fix legal regarding jp2 standard or will that just postpone (not for those issues though) the moment when a crash will happen ? Are the values legal ? I f not, could they be checked rather than just `continuing`.

trylab:
@mayeut The fix just solve the issues on the spot. I think there are two ways to solve them on the spot. 

If division-by-zero problems are detected, just ``continue;`` or ``return OPJ_FALSE;``.  Since ``return OPJ_TRUE;`` only happens in a single ``if`` statement, I think ``continue`` is more preferable.

The root cause needs further investigate.

mayeut:
All `opj_pi_next_*` shall probably be fixed (not all need a fix I think).
@detonin, I won't be able to investigate the root cause for this.
Two options:
- Merge this now & create an issue to identify the root cause.
- Wait for someone to identify the root cause & fix it
The second option is only to be considered if it can be done in the short term.

malaterre:
I vote for option (2), it seems we are not doing sanity checks on inputs. See for example https://github.com/uclouvain/openjpeg/issues/731#issuecomment-248862841

trylab:
Hi, I uploaded a minimized poc file at https://github.com/uclouvain/openjpeg/issues/731#issuecomment-248895070

trylab:
After I did some investigation on this issue, I think it's hard to do sanity checks earlier.

The following j2k image was taken from the **JPEG2000 Standard Document (Part 1)** (``ISO/IEC 15444-1``). It's a legal j2k image.
```
FF 4F FF 51 00 29 00 00 00 00 00 01 00 00 00 09
00 00 00 00 00 00 00 00 00 00 00 01 00 00 00 09
00 00 00 00 00 00 00 00 00 01 07 01 01 FF 5C 00
07 40 40 48 48 50 FF 52 00 0C 00 00 00 01 00 01
04 04 00 01 FF 90 00 0A 00 00 00 00 00 1E 00 01
FF 93 C7 D4 0C 01 8F 0D C8 75 5D C0 7C 21 80 0F
B1 76 FF D9
```

We can create a proof-of-concept file just by modifying three bytes.
```
FF 4F FF 51 00 29 00 00 00 00 00 01 00 00 00 09
00 00 00 00 00 00 00 00 00 00 00 01 00 00 00 09
00 00 00 00 00 00 00 00 00 01 07 01 [02] FF 5C 00     // #1
07 40 40 48 48 50 FF 52 00 0C 00 [04] 00 01 00 [11]    // #2  #3
04 04 00 01 FF 90 00 0A 00 00 00 00 00 1E 00 01
FF 93 C7 D4 0C 01 8F 0D C8 75 5D C0 7C 21 80 0F
B1 76 FF D9
```

1. Change ``l_img_comp->dy`` to ``0x02``, this byte will be read in function ``opj_j2k_read_siz``. The value will be propagated to ``comp->dy`` in function ``opj_pi_next_cprl``. Values in interval ``[1,255]`` are all valid.
2. Change ``SGcod(A)`` to ``0x04``, the value is assigned to the **Progression order** field which can be used to control which ``opj_pi_next_*`` function will be called. The valid values can be ``0,1,2,3,4``. Here ``4`` is used to make sure that function ``opj_pi_next_cprl`` will be called.
3. Change first byte of ``SPcod`` to ``0x11``, the value is assigned to ``l_tccp->numresolutions`` in function ``opj_j2k_read_SPCod_SPCoc``. Values in interval ``[0,32]`` are all valid. The value can be used to control ``levelno`` in function ``opj_pi_next_cprl``.
4. Here the value of ``Scod`` is ``0`` which can be used to make sure that values of ``res->pdx`` and ``res->pdy`` in function ``opj_pi_next_cprl`` are all ``0x0F``. The ``0x0F`` values are initialized in function ``opj_j2k_read_SPCod_SPCoc``. Also, the value is legal according to the standard document.

Now we can figure out why the division-by-zero problem will be happened. So I think that all the values are legal and it's hard to do sanity checks earlier.

BTW, I'm not familiar with the whole standard. The above text is just my analysis. Please point out if something is wrong. Thanks.

detonin:
@trylab thanks for the poc
@mayeut values are legal indeed but AFAIU, division by zero occurs because of an exceeding amount of left-shift operations ... So is it not OPJ_INT32 the problem here ?

malaterre:
@detonin I do not believe so since kakadu is also failing later in the pipeline (as indicated here: https://github.com/uclouvain/openjpeg/issues/731#issuecomment-248895070)

rouault:
I've fixed those issues in a somewhat equivalent way in https://github.com/uclouvain/openjpeg/commit/d27ccf01c68a31ad62b33d2dc1ba2bb1eeaafe7b . Closing

PR #866, opened 2016-11-06
fix static build only against tiff and its indirect dependencies

tSed:


ffontaine:
@rouault do you have some comments on this PR? It would be great if could be merged so we could remove it from buildroot.

rouault:
I'm not a CMake expert, but I'm wondering what is the effect on this change on dynamic builds

ffontaine:
This change will not break dynamic builds, it has been applied in buildroot for more than 4 years now.

You can find more information about pkg-config here: https://linux.die.net/man/1/pkg-config. Basically, with this change, the `Libs.Private` field of the pkg-config files (e.g. `libtiff-4.pc`) will be saved in `PC_TIFF_STATIC_LIBRARIES` if pkg-config is called with the `--static` argument (see https://cmake.org/cmake/help/v3.0/module/FindPkgConfig.html).

rouault:
ok thanks for the explanations. let's merge this

ffontaine:
Thanks, if you could also merge #867, it would be perfect. If you have any issues related to these these pull requests, do not hesitate to ping me.

Issue #863, opened 2016-11-02
Openjpeg-2.1.2 DoS vulnerability due to some logic error


twelveand0:
# Overview
I have found a vulnerability in openjpeg-2.1.2 (an open-source JPEG 2000 codec written in C
language) using AFL (http://lcamtuf.coredump.cx/afl/). The vulnerability exists in code responsible
for decoding the input image. The vulnerability is caused by an improper assumption: if the
decoding process successfully finishes, buffer of each component corresponding to each channel
(Red, Green, Blue, Alpha) has already been allocated and filled with data; however, the assumption
is not always valid, i.e. these buffers has not been allocated when the decoding process successfully
returns. The vulnerability can be viewed as a logic error. The vulnerability can trigger many different
crash points by crafting the PoC image file and cause Denial-of-Service due to Null Pointer
Reference. It’s probably that the vulnerability can cause crashes of some other type and cause
more critical impact by crafting the PoC.

# Analysis and PoC
The detail analysis and poc file can be found in the attachment.
[report2_+poc.zip](https://github.com/uclouvain/openjpeg/files/566799/report2_.poc.zip)

# Author
name: twelveand0 @ VARAS of IIE
email: l.bingchang.bc@gmail.com
org: IIE (http://iie.ac.cn)

# Notes
I have reported this to RedHat Security Team and they suggested me to report it here before assigning cve id.

attritionorg:
Dupe to #857?

twelveand0:
The bug can cause many different crash points where image->comps[i].data is accessed directly or indirectly after function "opj_decode" returns TRUE. #857 is one of them due to the same root cause analyzed in the report. The crash of #857 can be easily changed to by make the following check (line 1771-1780 in convert.c) to be FALSE.

```
1771       if ((force_split == 0) &&
1772				(ncomp == 2 /* GRAYA */
1773            || (ncomp > 2 /* RGB, RGBA */
1774               && image->comps[0].dx == image->comps[1].dx
1775                && image->comps[1].dx == image->comps[2].dx
1776                && image->comps[0].dy == image->comps[1].dy
1777                && image->comps[1].dy == image->comps[2].dy
1778                && image->comps[0].prec == image->comps[1].prec
1779                && image->comps[1].prec == image->comps[2].prec
1780                )))
1781		{
```
[...]

Issue #862, opened 2016-11-02
Openjpeg-2.1.2 Heap Buffer Overflow Vulnerability due to Insufficient check


twelveand0:
# overview
I have found a vulnerability in openjpeg-2.1.2 (an open-source JPEG 2000 codec written in C
language) using AFL (http://lcamtuf.coredump.cx/afl/). The vulnerability is a heap buffer overflow
vulnerability, which can cause out-of-bound read. The vulnerability exists in code responsible for
decompressing the input image. The vulnerability is caused by insufficient check: the code have
made sure the parameters of RED channel, GREEN channel and BLUE channel are the same,
however, does not do it for ALPHA channel, which causes the length of alpha buffer is less than the
expected access length. The vulnerability can cause Denial-of-Service and Information Disclosure
and I am not sure whether it can cause RCE.

# Analysis and PoC
The detail analysis and poc file can be found in the attachment.
[report1_+_poc.zip](https://github.com/uclouvain/openjpeg/files/566786/report1_._poc.zip)

# Author
name: twelveand0 @ VARAS of IIE
email: l.bingchang.bc@gmail.com
org: IIE (http://iie.ac.cn)

# Notes
I have reported this to RedHat Security Team and they suggested me to report it here before assigning cve id.

rouault:
Fixed per 2fa0fc61f2d546c8b67e7c5a9cbc61d98e1f7af0

PR #869, opened 2016-11-28
Changes for issues #863 and #862

szukw000:
This patch is reported to prevent SEGFAULTS introduced with files PoC1 and PoC2.


rouault:
Closing at superseded by #895

Issue #872, opened 2016-12-07
Out-of-Bounds Write issue can occur in function convert_32s_C1P1 (openjpeg-2.1.2/src/bin/jp2/convert.c:153)


chunibalon:
# DESCRIPTION

There is an out-of-bound write issue can occur in function convert_32s_C1P1 (openjpeg-2.1.2/src/bin/jp2/convert.c:153)(Actually, other functions like covert_*_* will also trigger this problem).
This issue can be caused by a malformed TIFF file.

# VERSION

OPENJPEG-2.1.2

# Address Sanitizer Output

==126239==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60200000ef51 at pc 0x7ffff6ef6935 bp 0x7fffffff7d30 sp 0x7fffffff74d8
READ of size 128 at 0x60200000ef51 thread T0
    #0 0x7ffff6ef6934 in __asan_memcpy (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x8c934)
    #1 0x40d325 in convert_32s_C1P1 /home/kirito/Desktop/fuzz/openjpeg-2.1.2/src/bin/jp2/convert.c:153
    #2 0x4466ce in tiftoimage /home/kirito/Desktop/fuzz/openjpeg-2.1.2/src/bin/jp2/converttif.c:1431
    #3 0x40b941 in main /home/kirito/Desktop/fuzz/openjpeg-2.1.2/src/bin/jp2/opj_compress.c:1739
    #4 0x7ffff5df982f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
    #5 0x4037f8 in _start (/home/kirito/Desktop/fuzz/openjpeg-2.1.2/build/asan-dbg-bin/tif_crashes/analysis/opj_compress+0x4037f8)

0x60200000ef51 is located 0 bytes to the right of 1-byte region [0x60200000ef50,0x60200000ef51)
allocated by thread T0 here:
    #0 0x7ffff6f02602 in malloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x98602)
    #1 0x446561 in tiftoimage /home/kirito/Desktop/fuzz/openjpeg-2.1.2/src/bin/jp2/converttif.c:1406
    #2 0x40b941 in main /home/kirito/Desktop/fuzz/openjpeg-2.1.2/src/bin/jp2/opj_compress.c:1739
[...]

Issue #871, opened 2016-12-07
Out-of-Bounds Write issue caused by integer overflow that can occur in function convert_8u32s_C1R()(openjpeg-2.1.2/src/bin/jp2/convert.c:368).


chunibalon:
# DESCRIPTION
There is an out-of-bound write issue caused by integer overflow that can occur in function convert_8u32s_C1R()(openjpeg-2.1.2/src/bin/jp2/convert.c:368).
This issue can be caused by a malformed TIFF file.

# VERSION
OPENJPEG-2.1.2

# Address Sanitizer Output
==61712==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x619000006380 at pc 0x00000040f586 bp 0x7fffffff7d20 sp 0x7fffffff7d10
READ of size 1 at 0x619000006380 thread T0
    #0 0x40f585 in convert_8u32s_C1R /home/kirito/Desktop/fuzz/openjpeg-2.1.2/src/bin/jp2/convert.c:368
    #1 0x4466ab in tiftoimage /home/kirito/Desktop/fuzz/openjpeg-2.1.2/src/bin/jp2/converttif.c:1430
    #2 0x40b941 in main /home/kirito/Desktop/fuzz/openjpeg-2.1.2/src/bin/jp2/opj_compress.c:1739
    #3 0x7ffff5df982f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
    #4 0x4037f8 in _start (/home/kirito/Desktop/fuzz/openjpeg-2.1.2/build/asan-dbg-bin/tif_crashes/analysis/opj_compress+0x4037f8)

0x619000006380 is located 0 bytes to the right of 1024-byte region [0x619000005f80,0x619000006380)
allocated by thread T0 here:
    #0 0x7ffff6f02602 in malloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x98602)
    #1 0x4464d1 in tiftoimage /home/kirito/Desktop/fuzz/openjpeg-2.1.2/src/bin/jp2/converttif.c:1399
    #2 0x40b941 in main /home/kirito/Desktop/fuzz/openjpeg-2.1.2/src/bin/jp2/opj_compress.c:1739
    #3 0x7ffff5df982f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)

[...]

Issue #811, opened 2016-08-18
fuzzing-created crashers


setharnold:
Hello; there's a renewed interest in adding openjpeg2 to Ubuntu's 'main' archive so that it can be used in Ubuntu's PDF reader, graphics programs, etc. As part of this process I've run the OpenJPEG 2.1.1 release through the AFL (American Fuzzy Lop) guided fuzzing tool and generated the following set of files that crash the opj_decompress tool. (Presumably this does something similar to what would be used by document viewers.)

The Ubuntu bug for the main inclusion process is at https://bugs.launchpad.net/ubuntu/+source/openjpeg2/+bug/711061

The crashing inputs are saved in a compressed tarball on launchpad: https://bugs.launchpad.net/ubuntu/+source/openjpeg2/+bug/711061/+attachment/4723094/+files/crashes-openjpeg-2.1.1.tar.gz

Please consider adding these images to the testing frameworks.

Thanks

malaterre:
Could you please also document the AFL cmd line you used ? In particular how long have you been runnning it ?

setharnold:
On Aug 17, 2016 22:57, "Mathieu Malaterre" <notifications@github.com> wrote:
>
> Could you please also document the AFL cmd line you used ? In particular
how long have you been runnning it ?
>

It was along the lines of:

screen
afl-fuzz -M master -i in -s out -f foo.j2k ../bin/opj_decompress -i @@ -o
foo.bmp
for i in `seq -w 1 29` ; do screen afl-fuzz -S slave$i -i in -s out -f
foo$i.j2k ../bin/opj_decompress -i @@ -o foo$i.bmp ; done

I ran this for about two days; afl-whatsup reported somewhere around two
months of CPU time. It was still finding new crashes every few minutes but
the weather is turning too hot to leave this running.

Change the 29 to something that's a good fit for your hardware. Roughly the
number of hyperthreads on your computer is a good starting point.

This returned roughly 4500 unique crashes. afl-cmin was able to reduce this
to a minimized set of roughly 600 input files that together cover all
"edges" covered by the full set. The afl-cmin command was along the lines
of:

mkdir collected min
mv out/*/*crash*/id:* collected/
afl-cmin -i collected/ -o min/ -f foo.j2k ../bin/opj_decompress -i @@ -o
foo.bmp

I started with the same relax.j2k image as last time. I think AFL could
make faster progress if it starts with a smaller input file. Next time
[...]

PR #875, opened 2016-12-15
Patch for crashes #811,#862,#863,#871 and #872

szukw000:
This patch supersedes patches #869 and #873


rouault:
Closing at superseded by #895

Issue #674, opened 2015-12-26
RESTART and BYPASS encoding modes are broken


boxerab:
i.e. lossless encoding with either mode produces a lossy encoded file

Simple test:
1. lossless encode with command line flag    `-M 4`
2. decode and compare with original :  non-zero MSE between two images

boxerab:
This is also true for BYPASS mode (M == 4)

boxerab:
These modes have been broken for at least 6 years, probably forever.

julienmalik:
Confirmed here.

By a strange coincidence, mode 1 and mode 4 are the only two modes who are generating the uninitialized read reported in #612. The others are valgrind-clean.

rouault:
Fixed per https://github.com/uclouvain/openjpeg/pull/949

rouault:
Fixed per https://github.com/uclouvain/openjpeg/pull/949

Issue #209, opened 2013-02-07
Issue using coder BYPASS (-M 1)


gcode-importer:
Originally reported on Google Code with ID 209

```
I am seeing artifacting when encoding with a self compiled OpenJPEG 1.5.1 on Debian.


Encoding ~4500x6000 greyscale tiffs with the following options results in artifacts
in the output jp2 files:
-I -p RPCL -n 6 -EPH -SOP -c [256,256],[256,256],[128,128],[128,128],[128,128],[128,128]
-b 64,64 -M 1 -r 320,160,80,40,20,11.25,7,4.6,3.4,2.75,2.4,1

When encoded with these settings all files have artifacts and one file cannot be opened
at all, despite jpylyzer reporting the headers as valid.  Changing the parameters to
-n 7 allows that file to be opened but it still shows the same artifacts as other files.

No artifacts are apparent when coder bypass is removed from the parameters.


```

Reported by `william.palmer@bl.uk` on 2013-02-07 14:30:50

rouault:
Fixed per https://github.com/uclouvain/openjpeg/pull/949

PR #894, opened 2017-02-15
fixed uninitialised memory access, mitigated BYPASS/RESTART issues

xbarin02:
cf. issues #612 #674 #209
This patch fixes uninitialised memory access caused by advancing a pointer in MQ coder. It also fixes BYPASS mode initialization and flush.


rouault:
Fixed per https://github.com/uclouvain/openjpeg/pull/949

Issue #911, opened 2017-04-26
Setup benchmark tool for OPJ speed evaluation


detonin:
Script to automatically assess speed and memory performances on a given test set and compare to kakadu

rouault:
Implemented per #917

PR #917, opened 2017-05-04
Add performance benchmarking scripts

rouault:
And run them by Travis-CI

Note: I've some doubt about the reproductibility and trustability in those tests on Travis infrastructure. Initially I had tried them on OSX workers too but this seemed far too unreliable : https://travis-ci.org/rouault/openjpeg/jobs/228744947

This relates to #911

### DIFF
--- a/tools/travis-ci/run.sh
+++ b/tools/travis-ci/run.sh
@@ -184,7 +184,9 @@ export OPJ_BINARY_DIR=$(opjpath -m ${PWD}/build)
 export OPJ_BUILD_CONFIGURATION=${OPJ_CI_BUILD_CONFIGURATION}
 export OPJ_DO_SUBMIT=${OPJ_DO_SUBMIT}
 
-ctest -S ${OPJ_SOURCE_DIR}/tools/ctest_scripts/travis-ci.cmake -V || true
+if [ "${OPJ_SKIP_REBUILD:-}" != "1" ]; then
+    ctest -S ${OPJ_SOURCE_DIR}/tools/ctest_scripts/travis-ci.cmake -V || true
+fi
 # ctest will exit with various error codes depending on version.
 # ignore ctest exit code & parse this ourselves
 set +x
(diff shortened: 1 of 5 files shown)


tools/travis-ci/run.sh
```diff
@@ -303,4 +305,43 @@ New/unknown test failure found!!!
...
+        echo "Comparing current version with Kakadu"
+        python compare_perfs.py /tmp/kakadu.csv /tmp/new.csv || true
+    fi
+    cd ..
+
+    REF_VERSION=master
+    if [ "${TRAVIS_PULL_REQUEST:-false}" == "false" ]; then
+        REF_VERSION=v2.1.2
+    fi
+    if [ ! -d ref_ojp ]; then
```
detonin: strangely enough, the correct prefix or suffix for OpenJPEG is opj and not ojp ... As everything in the code follows opj, I suggest we keep this rule.

tests/perf_test_filelist.csv
```diff
@@ -0,0 +1,7 @@
+filename,iterations,threads,command,comment
+../data/input/nonregression/kodak_2layers_lrcp.j2c,3,1,DECOMPRESS,
+../data/input/nonregression/kodak_2layers_lrcp.j2c,5,2,DECOMPRESS,
+../data/input/nonregression/kodak_2layers_lrcp.j2c,10,4,DECOMPRESS,
+../data/input/conformance/p0_07.j2k,3,1,DECOMPRESS,128x128 tiles RLCP 3decomp 8layers no MCT reversible
+../data/input/conformance/p0_04.j2k,10,1,DECOMPRESS,precincts 128x128 irreversible
+../data/input/nonregression/X_4_2K_24_185_CBR_WB_000.tif,3,1,COMPRESS,
```
detonin: do we need a newline here ?

rouault:
> move performance-related files (scripts and filenames list to a "performance" folder in the "tests" directory)

Done

rouault:
> instead of having the coding options hard-coded in the script, it would be useful to read an external file containing the options used in different environments

Agreed. I imagined that the COMPRESS string could be later extended to accept options like COMPRESS[reversible=yes,...] depending on what options we want to test.

tools/travis-ci/run.sh
rouault: Done

tests/perf_test_filelist.csv
rouault: Done (wasn't strictly needed, but better indeed)

rouault:
I've disabled making the job fail in case of performance regression, because even on Linux this is too unreliable. See https://travis-ci.org/uclouvain/openjpeg/jobs/229085228 that failed whereas previous builds (with same code) succeeded

detonin:
> Agreed. I imagined that the COMPRESS string could be later extended to accept options like COMPRESS[reversible=yes,...] depending on what options we want to test.

ok. This way is possible indeed but would require a "translation" from this generic syntax to the actual kdu or opj options I guess.

> I've disabled making the job fail in case of performance regression, because even on Linux this is too unreliable. See https://travis-ci.org/uclouvain/openjpeg/jobs/229085228 that failed whereas previous builds (with same code) succeeded

ok. I think you're right : benchmark results on Travis will never be trustful. I wonder if we should not even disable the test but provide an easy way to run the script locally, as you suggested. To be discussed.
Do you already have results or graphs showing the current performances of OPJ vs KDU ?

rouault:
> I wonder if we should not even disable the test but provide an easy way to run the script locally, as you suggested

The test can be run locally with
OPJ_SKIP_REBUILD=1 OPJ_CI_PERF_TESTS=1 OPJ_NONCOMMERCIAL=1 tools/travis-ci/run.sh

I think we can keep it on Travis in non-failing mode as I did in the latest version of the patch, so as a hint of possible regressions or improvements, to be confirmed locally.

> Do you already have results or graphs showing the current performances of OPJ vs KDU ?

Yes, extract from https://travis-ci.org/uclouvain/openjpeg/jobs/229431228
```
Comparing current version with Kakadu
../../data/input/nonregression/kodak_2layers_lrcp.j2c, 3 iterations, 1 threads, DECOMPRESS: ref_time 1830 ms, new_time 3196 ms, (regression) 74.6 %, ERROR_THRESHOLD
../../data/input/nonregression/kodak_2layers_lrcp.j2c, 5 iterations, 2 threads, DECOMPRESS: ref_time 1708 ms, new_time 4584 ms, (regression) 168.4 %, ERROR_THRESHOLD
../../data/input/nonregression/kodak_2layers_lrcp.j2c, 10 iterations, 4 threads, DECOMPRESS: ref_time 2138 ms, new_time 7831 ms, (regression) 266.3 %, ERROR_THRESHOLD
../../data/input/conformance/p0_07.j2k, 3 iterations, 1 threads, DECOMPRESS: ref_time 1948 ms, new_time 7418 ms, (regression) 280.8 %, ERROR_THRESHOLD
../../data/input/conformance/p0_04.j2k, 10 iterations, 1 threads, DECOMPRESS: ref_time 1096 ms, new_time 1270 ms, (regression) 15.9 %, ERROR_THRESHOLD
../../data/input/nonregression/X_4_2K_24_185_CBR_WB_000.tif, 3 iterations, 1 threads, COMPRESS: ref_time 4960 ms, new_time 9890 ms, (regression) 99.4 %, ERROR_THRESHOLD
TOTAL: ref_time 13683 ms, new_time 34192 ms, (regression) 149.9 %, ERROR_THRESHOLD
```

Issue #128, opened 2012-02-20
Uniformize code syntax 


gcode-importer:
Originally reported on Google Code with ID 128

```

see http://code.google.com/p/openjpeg/wiki/CodingStyle
```

Reported by [detonin](https://github.com/detonin) on 2012-02-20 16:22:14

stweil:
Indenting with tab characters (instead of spaces) is only an acceptable choice with a fixed tab width of 8 (that's the style used for Linux kernel code).

I personally prefer indentation with spaces, either in steps of 4 (preferred, used for many free software projects) or 2.

boxerab:
This tool might be useful here:  http://astyle.sourceforge.net/

an automatic formatter. You can specify the style you would like, and it will run through all source code files and re-format.

stweil:
Yes. I can confirm that astyle works pretty good.

detonin:
I gave astyle a try and pushed the result on the "codingstyle" branch, together with the config file I used (see tools/cstyle/opj_astyle.cfg in that branch). Plan is to merge it into master once we agree on the config file. Btw, is it better to do this just before the release or just after ?

Comments welcome.

There was also uncrustify proposed by @malaterre a while ago.
My opinion is that astyle works fine and seems simpler and lighter than uncrustify. Being simpler is an advantage if we want any contributor to process its PR through astyle before submitting.

@mayeut Do you think it would be possible to add a test in Travis to run astyle with the opj config file (might be in dry-run mode (--dry-run option)), and return a failure if there is any file being marked as changed ? The idea would be to merge PR only if it meets opj coding style.

mayeut:
My comments, all this is only a personal preference of course.
- I'm ok with k&r (c.f. minor comment below) or allman style. I usually mix them which is bad... (k&r for short blocks, allman for longer blocks . I find it more readable mixing the 2 for some reason.)
- If spaces are to be used, I'd rather see it in steps of 2. I find this easier to read, especially with many indentations levels where lines are either not fully readable (cropped by limited screen width or wrapped depending on preferences on this matter)
- with astyle, I'd probably add `--break-closing-brackets` if k&r style is used (the following is from style documentation:
```
void Foo(bool isFoo)
{
  if (isFoo) {
    bar();
  } else {
    anotherBar();
  }
}
becomes (a broken 'else'):

void Foo(bool isFoo) 
{
  if (isFoo) {
    bar();
  }
  else {
    anotherBar();
  }
}
```

[...]

PR #919, opened 2017-05-09
Add mechanisms to reformat and check code style, and reformat whole codebase (#128)

rouault:
Use an internal version of astyle (astyle 3.0). Scripts taken from QGIS.
astyle.options from https://github.com/uclouvain/openjpeg/issues/128
    
scripts/prepare-commit.sh can be used locally to automatically reformat edited files.
    
Travis-CI will run scripts/verify-indentation.sh to verify committed files.


rouault:
Example of (voluntary) failure of a Travis job due to syntax error: https://travis-ci.org/rouault/openjpeg/jobs/230491462

stweil:
While adding GPL code to `thirdparty` might be considered to be legally fine, I don't think that adding such code to `scripts` is a good idea.

rouault:
@detonin If you intend to do a release, I guess you might want to make it before I merge this.

rouault:
> While adding GPL code to thirdparty might be considered to be legally fine,
Actually the astyle source code is a X/MIT licence : https://github.com/rouault/openjpeg/blob/d4e54e9f35d532062533f1d369c159810b01d224/thirdparty/astyle/LICENSE.md

> I don't think that adding such code to scripts is a good idea.
Someone could possibly later rewrite them if they find that the investment is worth, but my shell abilities are too poor and this would be wheel reinvention. Anyway anybody can safely strip the scripts/ directory for a tarball and still successfully build openjpeg if they are so afraid of GPL.

boxerab:
@stweil packaging BSD and GPL files together doesn't trigger the GPL on the BSD project. It is only when the BSD proj is linked to the GPL that the GPL license is triggered. So, there shouldn't be any legal issues with putting a GPL script in the project.

scripts/astyle.options
```diff
@@ -0,0 +1,11 @@
+--convert-tabs
+--lineend=linux
+--indent=spaces=4
+--style=kr
+--add-brackets
```
detonin: should be --add-braces (--add-brackets is deprecated)

scripts/astyle.options
```diff
@@ -0,0 +1,11 @@
...
+--max-code-length=80
+--break-after-logical
+--pad-header
+--pad-oper
+--unpad-paren
+--suffix=none
```
detonin: @rouault Where does theses options come from ? Are they copy-pasted from QGIS ?
@malaterre @mayeut Any comment on the astyle config before we merge the PR ?

rouault: All those options come from https://github.com/uclouvain/openjpeg/issues/128#issuecomment-175076907

scripts/astyle.options
rouault: Fixed

.travis.yml
```diff
@@ -1,25 +1,37 @@
-language: c
+language: cpp
```
detonin: Is that required ? Changing the whole project from C to C++ (also in cmake) just for opjstyle seems a bit overkill. No other option ?

rouault: I found hard with Travis to make it pass on all configurations and I'm afraid I didn't find a better solution.

detonin: Are there any implication of changing compiler from gcc to g++ ? Does this prevent to compile on some older platforms ? 
@malaterre any advice on this ?

detonin:
> @detonin If you intend to do a release, I guess you might want to make it before I merge this.

I'll rather do the release afterwards. But before merging I would like some feedback on the compiler change.

@boxerab thanks for your comment on licensing

.travis.yml
rouault: cmake will require a c++ compiler to be detected, even if it doesn't use it (it is only used if WITH_ASTYLE is ON, which is not the default). There's perhaps a way to restrict c++ compiler detection only when WITH_ASTYLE is ON but I couldn't find it.
The C compiler (gcc) will still be used to compile the .c files.

Anyway we require CMake to be able to build openjpeg and CMake source code itself is C++, so the platform must have a working C++ compiler.

And at some point we might perhaps want to use C++ for openjpeg itself.

malaterre: @rouault you are describing the symptoms not the real issue :) The fact that cmake only works on arch with valid c++ compiler is outside the scope of this question. I also fail to understand why `WITH-ASTYLE=ON` imply a valid c++ compiler in the PATH. I do understand that it should have zero impact on the build itself (a bit more dependencies though).

rouault: @malaterre Astyle source code is C++ and apparently the fact that a project is c, c++ or both is declared in cmake at the PROJECT() scope, so as soon as you have C++ code you must enable C++. Is it really an issue to require a C++ compiler available ? I don't think so honestly.

malaterre: I totally failed to follow the fact that now openjpeg is bootstraping a c++ project (astyle). In that case, yes there isn't much solution. As said above `it should have zero impact in the build itself`, I was just curious.

detonin: ok thanks for your feedback @malaterre .

@rouault I think we can proceed and merge now.

rouault:
@detonin Reformat source code PR merged

rouault:
@detonin Wait: I see the .h files were not reformatted. Addressed by https://github.com/uclouvain/openjpeg/pull/926

rouault:
@detonin OK, #926 merged. Things should be good now

PR #954, opened 2017-06-16
build both shared and static library

jeroen:
This enabled building a static library, in addition to a shared one. I need this myself for building standalone applications, but I think it is generally useful.

Thank you for your review.


src/lib/openjp2/CMakeLists.txt
```diff
@@ -87,7 +87,9 @@ if(WIN32)
     add_definitions(-DOPJ_STATIC)
   endif()
 endif()
-add_library(${OPENJPEG_LIBRARY_NAME} ${OPENJPEG_SRCS})
+add_library(${OPENJPEG_LIBRARY_NAME} SHARED ${OPENJPEG_SRCS})
+add_library(openjp2_static STATIC ${OPENJPEG_SRCS})
```
rouault: Any reason not use ${OPENJPEG_LIBRARY_NAME} as well ? Generally both static and dynamic library have the same basename.

jeroen: cmake doesn't allow for duplicate target name I think

rouault: https://stackoverflow.com/questions/2152077/is-it-possible-to-get-cmake-to-build-both-a-static-and-shared-version-of-the-sam and https://datainfer.wordpress.com/2013/10/24/make-both-static-and-shared-libraries-in-one-build-with-cmake/ give one hint how to do that with set_target_properties(MyLibStatic PROPERTIES OUTPUT_NAME MyLib), but this one must be restricted to non-Windows platform, as Windows apparently need different names.

jeroen: Yes the PR already uses the `set_target_properties`.

rouault:
Fails to build on Windows : http://my.cdash.org/buildSummary.php?buildid=1229211

jeroen:
OK let me condition this on Non windows.

jeroen:
Thanks, can you try again?

jeroen:
Ping? This should be good to go. Thanks for reviewing!

PR #956, opened 2017-06-18
Unknown input file format error for .tiff extension

Jamaika1:
opj_compress gives Unknown input file format error for .tiff extension  #950


GitHub2Janky (commented):
#

PR #968, opened 2017-07-06
Reduce memory decoding

rouault:
On MAPA_002.jp2, peak memory goes from 4.5 GB to 1.9 GB (including the 1.6 GB needed for the whole image)
The improvements consist of two main independant changes:
* the major one (4.5 GB->2.7 GB): in case of decoding of whole image consisting of a single time, use directly the tile component data as output image data. And save the use of the temporary buffer to transfer between both
* a more modest one (2.7->1.9GB) : avoid allocating memory to store the part of the codestream that is dedicated to a buffer. But instead point to the codestream buffer itself.

The major one has an annoying side effect: given that tile component data buffers are 16-byte aligned (for SSE code paths), we need in all code paths that alloc/free image->comps[].data to use aligned malloc/free as well (on Linux, with posix_memalign() this isn't visible since such buffers are free'd by free() too, but on Windows mismatch of allocator/deallocator is crashing). This might be seen as a slight API incompatibility in case someone would manually allocate/free the buffers of a opj_image (but nothing in the API indicates that this is something people are actually allowed to do and what the allocators/freeors are supposed to be), which for standard decoding/encoding code is not needed.

### DIFF
--- a/src/bin/common/color.c
+++ b/src/bin/common/color.c
@@ -122,9 +122,9 @@ static void sycc444_to_rgb(opj_image_t *img)
     cb = img->comps[1].data;
     cr = img->comps[2].data;
 
-    d0 = r = (int*)malloc(sizeof(int) * max);
-    d1 = g = (int*)malloc(sizeof(int) * max);
-    d2 = b = (int*)malloc(sizeof(int) * max);
+    d0 = r = (int*)opj_image_data_alloc(sizeof(int) * max);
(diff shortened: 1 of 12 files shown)


rouault:
Community probing regarding the change for allocation components of opj_image was done in this email thread : https://groups.google.com/forum/#!topic/openjpeg/RoG0gICchsk

PR #975, opened 2017-07-31
Catch images broken by AFL

szukw000:
@rouault ,
the minimal archive is attached.
winfried
[openjpeg-crashes-2017-07-30.zip](https://github.com/uclouvain/openjpeg/files/1187460/openjpeg-crashes-2017-07-30.zip)

### DIFF
--- a/src/bin/jp2/opj_dump.c
+++ b/src/bin/jp2/opj_dump.c
@@ -577,6 +577,8 @@ int main(int argc, char *argv[])
         opj_set_warning_handler(l_codec, warning_callback, 00);
         opj_set_error_handler(l_codec, error_callback, 00);
 
+        parameters.dump_state = 1; /* AFL test */
+
         /* Setup the decoder decoding parameters using user parameters */
         if (!opj_setup_decoder(l_codec, &parameters)) {
             fprintf(stderr, "ERROR -> opj_dump: failed to setup the decoder\n");
--- a/src/bin/jp2/opj_dump.c
+++ b/src/bin/jp2/opj_dump.c
@@ -577,7 +577,7 @@ int main(int argc, char *argv[])
         opj_set_warning_handler(l_codec, warning_callback, 00);
         opj_set_error_handler(l_codec, error_callback, 00);
 
-        parameters.dump_state = 1; /* AFL test */
+        parameters.flags |= OPJ_DPARAMETERS_DUMP_FLAG;
 
         /* Setup the decoder decoding parameters using user parameters */
         if (!opj_setup_decoder(l_codec, &parameters)) {
--- a/src/lib/openjp2/j2k.c
+++ b/src/lib/openjp2/j2k.c
@@ -2054,6 +2054,7 @@ static OPJ_BOOL opj_j2k_read_siz(opj_j2k_t *p_j2k,
     OPJ_UINT32 l_remaining_size;
     OPJ_UINT32 l_nb_tiles;
     OPJ_UINT32 l_tmp, l_tx1, l_ty1;
+    OPJ_UINT32 l_prec0, l_sgnd0;
     opj_image_t *l_image = 00;
     opj_cp_t *l_cp = 00;
     opj_image_comp_t * l_img_comp = 00;
@@ -2156,7 +2157,20 @@ static OPJ_BOOL opj_j2k_read_siz(opj_j2k_t *p_j2k,
                       "Error with SIZ marker: illegal tile offset\n");
         return OPJ_FALSE;
     }
+    if (!p_j2k->dump_state) {
+        OPJ_UINT32 siz_w, siz_h; /* AFL test */
 
+        siz_w = l_image->x1 - l_image->x0;
+        siz_h = l_image->y1 - l_image->y0;
+
+        if (p_j2k->ihdr_w > 0 && p_j2k->ihdr_h > 0
+                && (p_j2k->ihdr_w != siz_w || p_j2k->ihdr_h != siz_h)) {
+            opj_event_msg(p_manager, EVT_ERROR,
+                          "Error with SIZ marker: IHDR w(%u) h(%u) vs. SIZ w(%u) h(%u)\n", p_j2k->ihdr_w,
+                          p_j2k->ihdr_h, siz_w, siz_h);
+            return OPJ_FALSE;
+        }
+    }
 #ifdef USE_JPWL
     if (l_cp->correct) {
         /* if JPWL is on, we check whether TX errors have damaged
@@ -2211,6 +2225,8 @@ static OPJ_BOOL opj_j2k_read_siz(opj_j2k_t *p_j2k,
 
     l_img_comp = l_image->comps;
 
+    l_prec0 = 0;
+    l_sgnd0 = 0;
     /* Read the component information */
     for (i = 0; i < l_image->numcomps; ++i) {
         OPJ_UINT32 tmp;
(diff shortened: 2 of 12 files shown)


szukw000:
@rouault ,
I have called 'cmake -DWITH_ASTYLE=on ..', then called 'make', then - from the BUILD directory -
called '../ scripts/prepare-commit.sh'.
winfried

rouault:
I can see that there are new test failures. Do you know if they are regressions or issues that are actually fixed ?
```
NR-DEC-issue414.jp2-110-decode

NR-DEC-issue414.jp2-110-decode-md5

NR-DEC-issue458.jp2-120-decode

NR-DEC-issue458.jp2-120-decode-md5
```

szukw000:
@rouault ,
issue458.jp2: Csiz(4)

comp sign prec dx  dy
```
0     0    5   1   1
1     0    5   1   1
2     0    5   1   1
3     0    1   1   1
```

issue414.jp2: Csiz(5)
```
comp sign prec dx  dy

0     0    8   1   1
1     0    8   1   1
2     0    8   1   1
3     0    8   1   1
4     0    8   1   1
```
Changed code in j2k.c, line 2341:

if (l_image->numcomps > 3 //== 4 /* RGBA */

winfried

rouault:
> Changed code in j2k.c, line 2341:

I don't think all the checks you added in j2k.c related to color space and number of channels should be put there. The standard doesn't mention them (as far as I can see  in "I.5.3.3 Colour Specification box"), so I guess we should accept images with apparently inconsistant channel number vs color space. We can reject them at the conversion stage, but in the library itself, we should probably allow such weird images, ie the color space is just a hint

src/bin/jp2/opj_dump.c
```diff
@@ -577,6 +577,8 @@ int main(int argc, char *argv[])
         opj_set_warning_handler(l_codec, warning_callback, 00);
         opj_set_error_handler(l_codec, error_callback, 00);
 
+        parameters.dump_state = 1; /* AFL test */
```
rouault: AFL test comment to remove and in other locations

src/lib/openjp2/j2k.c
```diff
@@ -2276,6 +2305,122 @@ static OPJ_BOOL opj_j2k_read_siz(opj_j2k_t *p_j2k,
             l_cp->m_specific_param.m_dec.m_reduce; /* reducing factor per component */
         ++l_img_comp;
     }
+    if (!p_j2k->dump_state) {
```
rouault: As discussed in another comment, I don't think those checks should be put in the library code

src/lib/openjp2/openjpeg.h
```diff
@@ -564,6 +564,7 @@ typedef struct opj_dparameters {
     /* <<UniPG */
 
     unsigned int flags;
+    unsigned int dump_state; /* AFL test */
```
rouault: This will change the ABI. An alternative way would be to add a OPJ_DPARAMETERS_DUMP_FLAG 0x0002 flag (see existing OPJ_DPARAMETERS_IGNORE_PCLR_CMAP_CDEF_FLAG)

rouault:
(note in case you're not familiar with github & pull requests : you can add new commits to your  changes-for-afl-tests branch and push it, and they will be added to this pull request)

szukw000:
Yes, I am not familiar with github. As I am not familiar I do not know

1. howto add a new commit

2. howto push it

winfried

> Even Rouault <notifications@github.com> hat am 31. Juli 2017 um 22:13 geschrieben:
> 
> 
>     (note in case you're not familiar with github & pull requests : you can add new commits to your changes-for-afl-tests branch and push it, and they will be added to this pull request)
>

rouault:
@winfried Do changes in your local tree and git commit them as usual, and then git push your branch to your github account as you did for the initial pull request. github will realize that your branch was the source of this pull request, so it will queue the new commits to it

rouault:
@szukw000 2 issues:
* There's a style issue (see https://travis-ci.org/uclouvain/openjpeg/jobs/260252276) that you can fix by running "scripts/astyle.sh src/bin/jp2/converttif.c"
* There's a compilation issue related to " error: unknown type name ‘tmsize_t’" on one target that haveOPJ_CI_PROFILE=1 and do a run with "cmake "-DCMAKE_C_FLAGS=-pg -O3" -DCMAKE_EXE_LINKER_FLAGS=-pg -DCMAKE_SHARED_LINKER_FLAGS=-pg -DBUILD_SHARED_LIBS=OFF .." ( see end of tools/travis-ci/run.sh). From the log, the system libtiff is used (instead of the one in third_party) and it must be a libtiff 3.X version that doesn't have tmsize_t.  It would be good if openjpeg could still build against libtiff 3.X. It doesn't look like your change of tsize_t to tmsize_t is needed since in libtiff 4.X both types are equivalent because of "typedef tmsize_t tsize_t" in tiffio.h.  See https://travis-ci.org/uclouvain/openjpeg/jobs/260252281 for full log

szukw000:
@rouault,

libopenjpeg uses libtiff-4.0.6 in 'thirdparty/libtiff'.
I suppose there is a problem with libtiff < 4.x.y on 64-bit machines:

tiff-3.8.0/test:
./ascii_tag
sizeof TSIZE_T(4) UINT_MAX(4294967295l)

tiff-4.0.7/test:
./ascii_tag
sizeof TSIZE_T(8) UINT_MAX(4294967295l)

rouault:
> I suppose there is a problem with libtiff < 4.x.y on 64-bit machines:

Not really, on libtiff 3.X, that's per design: "typedef int32 tsize_t;          /* i/o size in bytes */" .

Why did your change to tmsize_t was supposed to fix ?

szukw000:
@rouault,

I know:

3.8.0:
============
#define TIFFLIB_VERSION 20051230
* NB: tsize_t is int32 and not uint32 because some functions
* return -1.
<TD>typedef int32 tsize_t;</TD> <TD>i/o size in bytes</TD>


But an int32 type can not contain an UINT_MAX value but an INT_MAX value (2147483647) only.

winfried

rouault:
> But an int32 type can not contain an UINT_MAX value but an INT_MAX value (2147483647) only.

I agree, but I'm lost at what you are suggesting to do. What if we revert back to using tsize_t ? I'd say that even if there are some potential vulnerability with libtiff 3.X due to using its tsize_t, it is somehow OK. What matters is that we work fine with libtiff 4.X (IMHO)

szukw000:
@rouault,

some tests.

```

1. 32bit, linux-4.7.5, gcc-5.3.0, libtiff-4.0.6 :
converttif.c:imagetotif: TMSIZE_T(4) TSIZE_T(4) INT64(8) UINT_MAX(4294967295)
583: strip_size = UINT_MAX;

openjpeg2-2017-08-06/src/bin/jp2/converttif.c:583:15: warning: conversion of unsigned constant value to negative integer [-Wsign-conversion]
strip_size = UINT_MAX;

2. 64bit, WIN7, VS Community 2017, libtiff-4.0.6:
converttif.c:582: TSIZE_T(8) TMSIZE_T(8) UINT_MAX(4294967295)


3. 64bit, linux-4.12.3, gcc-7.1.0, libtiff-4.0.8:
converttif.c:581: INT(4) TMSIZE_T(8) INT64(8)

```

Ignoring the fact that travis does use old buggy libraries one could replace 'tsize_t' or 'tmsize_t'

with 'int64_t'. 'int64_t' can hold an UINT_MAX value.

winfried

rouault:
> one could replace 'tsize_t' or 'tmsize_t' with 'int64_t'

Sounds good

szukw000:
@rouault ,
I have now changed 'src/bin/jp2/converttif.c'. The changes look clumsy. I have tested
the file with linux32bit and linux64bit.

winfried

rouault:
@detonin The changes looks good to me. I let you look at them

detonin:
Looks good indeed. Merging.
@szukw000 Thanks for the contribution !

PR #990, opened 2017-08-16
Sub-tile decoding: only decode precincts and codeblocks that intersect the window specified in opj_set_decode_area()

rouault:
bin/opj_decompress -i ../MAPA.jp2 -o out.tif -d 0,0,256,256 runs in 2 s with this change vs 38.4 s with master


rouault:
Closing. Superseded by #1001

Issue #1014, opened 2017-09-08
Use opj_image_data_free instead of opj_free for image->comps[].data


trylab:
Currently, mismatches have been found in the following files:
+ ``src/lib/openjp2/image.c``
+ ``src/lib/openjp2/jp2.c``
+ ``src/lib/openmj2/jp2.c``

This may lead to crash on Windows as described in https://github.com/uclouvain/openjpeg/commit/61fb5dd7f81c2e3dfabbb99f59dc89572d59fa37

BTW, I don't know if **openjp3d**, **openjpip**, or **openjpwl** have same problems.
My patch could address issues in **openjp2** and **openmj2**.

rouault:
I left apart the openmj2 case since this lib wasn't really touched. It has currently a dummy opj_image_data_malloc/free implementation that just redirects to standard malloc/free

PR #1016, opened 2017-09-08
Fix crash on Windows by using opj_image_data_free instead of opj_free

trylab:
Fix crash on Windows by using ``opj_image_data_free`` instead of ``opj_free`` for ``image->comps[].data``
Issue: (https://github.com/uclouvain/openjpeg/issues/1014)


rouault:
Fixed per https://github.com/uclouvain/openjpeg/commit/b73ce715d2a484d7355639d863d0418a0e5b8858

PR #1022, opened 2017-09-19
Add capability to decode only a subset of all components of an image.

rouault:
#This adds a opj_set_decoded_components(opj_codec_t *p_codec,
OPJ_UINT32 numcomps, const OPJ_UINT32* comps_indices, OPJ_BOOL apply_color_transforms) function, and equivalent "opj_decompress -c compno[,compno]*" option.

When specified, neither the MCT transform nor JP2 channel transformations
will be applied (apply_color_transforms must be set to OPJ_FALSE)

Tests added for various combinations of whole image vs tiled-based decoding,
full or reduced resolution, use of decode area or not.

### DIFF
--- a/src/bin/jp2/opj_decompress.c
+++ b/src/bin/jp2/opj_decompress.c
@@ -152,6 +152,10 @@ typedef struct opj_decompress_params {
     int num_threads;
     /* Quiet */
     int quiet;
+    /** number of components to decode */
+    OPJ_UINT32 numcomps;
+    /** indices of components to decode */
+    OPJ_UINT32* comps_indices;
 } opj_decompress_parameters;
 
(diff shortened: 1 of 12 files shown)


rouault:
@detonin This is the functionnality I mentionned a few weeks ago during one of our meetings. I finally decided to add a new function since that enables more precise runtime checks (and the user can only know the total number of components after opening the file). Do you think this should be merged before or after the next 2.2.1 release ?

rouault:
@detonin Seing your email to @malaterre , I think you refered to the new function of this PR. Tell me if I should merge it, or just press the merge PR button if you're happy with it.

detonin:
@rouault yes indeed. Many thanks for this contribution, much appreciated. Merging now.

PR #1027, opened 2017-09-23
CMakeLists.txt: Don't require a C++ compiler

jacmet:
By default, CMake assumes that the project is using both C and C++.  By
explicitly passing 'C' as argument of the project() macro, we tell CMake
that only C is used, which prevents CMake from erroring out if a C++
compiler doesn't exist.

Signed-off-by: Peter Korsgaard <peter@korsgaard.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@ endif()
 #string(TOLOWER ${OPENJPEG_NAMESPACE} OPENJPEG_LIBRARY_NAME)
 set(OPENJPEG_LIBRARY_NAME openjp2)
 
-project(${OPENJPEG_NAMESPACE})
+project(${OPENJPEG_NAMESPACE} C)
 
 # Do full dependency headers.
 include_regular_expression("^.*$")


rouault:
This will break when using WITH_ASTYLE=ON which requires C++ to compile the (tooling) astyle binary. It would be good that CMake would only require a C compiler if compiling with WITH_ASTYLE=OFF, but I don't know how to do that

jacmet:
Hmm, true. I'm afraid I don't know enough about CMake to know how to handle that. I just noticed the build errors after we bumped the openjpeg version in Buildroot, E.G. http://autobuild.buildroot.net/results/e2f/e2ff0a7fa2b911157edf6c43a8eed797b22edd46/build-end.log

Considering the OpenJPEG tagline (an open-source JPEG 2000 codec written in C language), requiring a C++ compiler seems kind of silly :)

ahoffer:
@jacmet Thank you! I was unable to build openjpeg until I used the changes in your PR. The Makefile generated by CMake was missing most of the targets until I added "C".

rouault:
Closing. The proposed solution isn't appropriate

PR #1063, opened 2018-02-12
In case of an error in opj_decompress, the output file is deleted only if it's a regular file

adolski:
I need to stream output from `opj_decompress` by invoking it like:
```
ln -s /dev/stdout stdout.bmp
opj_decompres -i image.jp2 -o stdout.bmp > output.bmp
```
This patch will prevent `opj_decompress` from deleting symlinks or any other non-regular files in the event of an error.


rouault:
This breaks the Windows build and I'm not sure to really understand why a special case should be made on symlinks

PR #1090, opened 2018-02-24
Convert files to UTF-8 encoding

stweil:


rouault:
Could you drop the changes in wx ? Since that part of the code didn't go through reformatting and the reformatter check is confused

stweil:
Done. I rebased the PR and removed the wx commit.

stweil:
Should I send the wx commit separately, or how can that part be cleaned?

PR #1095, opened 2018-03-02
opj_mj2_extract: Check provided output prefix for length

### LINKED ISSUE #1088: sprintf buffer overflow
https://github.com/uclouvain/openjpeg/blob/06f7d412435ddb62f8e9935af7e306783bc4b75b/src/bin/mj2/opj_mj2_extract.c#L143

Hello, it looks like this `sprintf(3)` could overflow the buffer that is supplied for it.

Thanks

kbabioch:
This uses snprintf() with correct buffer length instead of sprintf(), which
prevents a buffer overflow when providing a long output prefix. Furthermore
the program exits with an error when the provided output prefix is too long.

Fixes #1088.

### DIFF
--- a/src/bin/mj2/opj_mj2_extract.c
+++ b/src/bin/mj2/opj_mj2_extract.c
@@ -140,7 +140,12 @@ int main(int argc, char *argv[])
         fread(frame_codestream, sample->sample_size - 8, 1,
               file); /* Assuming that jp and ftyp markers size do*/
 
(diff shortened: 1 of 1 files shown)


rouault:
This breaks formatting rules. Please build with WITH_ASTYLE=ON and run scripts/prepare-commit.sh

rouault:
Merged manually with formatting fix

PR #1097, opened 2018-03-02
Add sanitizer support

kbabioch:
This make is easy to compile OpenJPEG with sanitizer support. It relies on [arsenm/sanitizers-cmake](https://github.com/arsenm/sanitizers-cmake.git) and adds sanitizer support for several executables as well as all of the libraries.


kbabioch:
It includes [arsenm/sanitizers-cmake](https://github.com/arsenm/sanitizers-cmake.git) as submodule within Git. Let me know if for some reason you don't like this, we can also copy over the files, etc.

rouault:
Breaks Windows build
Not sure if we want to add git submodules

PR #1098, opened 2018-03-02
Fix an out-of-bounds read in imagetopgx()

### LINKED ISSUE #1069: out of bounds read
https://github.com/uclouvain/openjpeg/blob/da5e897232ef824daf9a492e746ed22cf2a43f18/src/bin/jp2/convert.c#L1375

Hello,

if `outfile` is shorter than four chars long, this will read out of bounds.

Thanks

kbabioch:
In case this function is invoked with a short outfile, a out-of-bounds read
will occur. This is currently not possible with opj_decompress, as this case
is intercepted in `get_file_format()` in `src/bin/jp2/opj_decompress.c`,
nevertheless it is not nice and robost code.

This fixes #1069.


kbabioch:
This needs some further refactoring, will re-submit.

PR #1104, opened 2018-03-04
Fix Mac builds

rouault:


rouault:
Kind of mysterious why changing from quiet to verbose mode makes thing work, but oh well...

PR #692, opened 2016-01-09
Add JPIP options to Travis configuration

stweil:

PR #1119, opened 2018-06-18
Use local type declaration for POSIX standard type only for MS compiler

stweil:
ssize_t is a POSIX type which is declared in POSIX include files.
Mingw-w64 provides it also for Windows.

Use the local declaration only with MS compilers.

### DIFF
--- a/src/lib/openjpip/sock_manager.c
+++ b/src/lib/openjpip/sock_manager.c
@@ -30,7 +30,9 @@
 
 #ifdef _WIN32
 #include <windows.h>
+#ifdef _MSC_VER
 typedef SSIZE_T ssize_t;
+#endif
 #else
 #include <sys/types.h>
 #include <sys/socket.h>


vtorri:
well, are you sure that vc++/Visual Studio defines ssize_t ? I know that it defines SSIZE_T, but I do not find ssize_t

i would have done : 
``
#ifdef _MSC_VER 
typedef SSIZE_T ssize_t;
#endif 
``

stweil:
Thank you for your hint. I only have Mingw-w64 installed locally, so could not test it with VS, but MSDN confirms what you said. I updated the commit as suggested.

stweil:
Updated subject line and description of PR to match updated commit.

stweil:
It looks like the continuous integration test still does not check the JPIP part of the code. See pull request #692 which tried to fix that some years ago.

PR #1138, opened 2018-08-29
Some changes for PNM file

szukw000:
See issue 1137


szukw000:
@rouault ,
this change proposal should be useful for the bug report of issue 137.
winfried

Issue #1137, opened 2018-08-29
Slow to compress certain crafted pgm image to jp2


HongxuChen:
When `opj_compress` convert certain pgm image to jp2, it may unexpectedly spend too much time. The corresponding code should lie at
https://github.com/uclouvain/openjpeg/blob/master/src/bin/jp2/convert.c#L1896-L1906

tested input file (please unzip firstly):
[test.pgm.zip](https://github.com/uclouvain/openjpeg/files/2331238/test.pgm.zip)

![2018-08-29-155406_569x181_scrot](https://user-images.githubusercontent.com/843267/44773698-c8b31a80-aba3-11e8-95d1-4a83bfbf4565.png)

szukw000:
@rouault ,
I have added #1141 to fix this.
winfried

PR #1141, opened 2018-08-31
Changes in pnmtoimage if image data are missing

szukw000:
Compare #1137


szukw000:
@rouault ,
this fixes #1137 . The test file does have header data but no image data.
winfried

PR #1142, opened 2018-09-05
wx: Convert ISO-8859 to UTF-8

stweil:


rouault:
Can you add an exception in scripts/astyle.sh for files in src/bin/wx/OPJViewer/source so they are not checked for formatting ? (look for "thirdparty" in astyle.sh for such an exception)

Issue #1057, opened 2018-01-19
Out-of-bound left shift in opj_j2k_setup_encoder (src/lib/openjp2/j2k.c)


ProbeFuzzer:
(This problem is discovered with UBSAN enabled)
On latest version (2.3) and master branch of openjpeg:
there is an integer overflow caused by out-of-bound left shift in opj_j2k_setup_encoder function (src/lib/openjp2/j2k.c), which could cause denial of service via a crafted bmp file.

src/lib/openjp2/j2k.c:7304:48: runtime error: shift exponent 4294967295 is too large for 32-bit type 'int'

To reproduce this issue, run: bin/opj_compress -n 1 -i $POC -o OUTPUT

The POC could be downloaded at: https://github.com/ProbeFuzzer/poc/blob/master/openjpeg/openjpeg_2-3_opj_compress_integer-overflow_opj_j2k_setup_encoder.bmp

carnil:
This was assigned [CVE-2018-5785](http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-5785)

hlef:
The underlying issue is actually in `src/bin/jp2/convertbmp.c`: BMP files declaring compression = 3 (`BI_BITFIELDS`) with header size <= 56 keep their color mask values to zero. This is not an issue for BMP files with bit count 16 because of a check added in 16240e21f9f51905d2f5b3b6a348d7a91b35497c which sets default values to the color masks if they are all 0. We should maybe have something like that for other bit counts as well. Also, there might be some similar issues in the case where header size >= 56 + some masks are intentionally set to zero.

I'll investigate further and PR a patch.

hlef:
Just submitted a PR addressing this issue.

tmdahr1245:
Hello.

I would like to tell you about CVE-2018-5785 addressed in this issue.

I found this vulnerability in version 2.1.2, and I also confirmed that integer overflow occurs in that version.
When reproducing the exploit, the PoC above was used as it is and the reproduction process is the same.

We would appreciate it if you confirm the information and reflect it.
Thanks.

PR #1148, opened 2018-09-22
CVE-2018-5785: fix issues with zero bitmasks

hlef:
```
In the case where a BMP file declares compression 3 (BI_BITFIELDS) with header size <= 56, all bitmask values keep their initialization value 0. This may lead to various undefined behavior later e.g. when doing 1 << (l_comp->prec - 1).

This issue does not affect files with bit count 16 because of a check added in 16240e2 which sets default values to the color masks if they are all 0.

This patch adds similar checks for the 32 bit case.

Also, if a BMP file declares compression 3 with header size >= 56 and intentional 0 bitmasks, the same issue will be triggered in both the 16 and 32 bit count case.

This patch adds checks to bmp_read_info_header() rejecting BMP files with "intentional" 0 bitmasks. These checks might be removed in the future when proper handling of zero bitmasks will be available in openjpeg2.
```

A few comments on the checks added in bmp_read_info_header():
* IMO the best solution would be to correctly handle zero bitmasks in openjpeg2. I tried to do it, but the code seems to widely assume prec > 0 which makes it too time consuming for me.
* 0 bitmasks are an obscure part of the BMP spec, so these checks are not fundamentally wrong. I consider them to be an acceptable temporary solution.

This PR addresses #1057 (CVE-2018-5785).

### DIFF
--- a/src/bin/jp2/convertbmp.c
+++ b/src/bin/jp2/convertbmp.c
@@ -435,15 +435,30 @@ static OPJ_BOOL bmp_read_info_header(FILE* IN, OPJ_BITMAPINFOHEADER* header)
         header->biRedMask |= (OPJ_UINT32)getc(IN) << 16;
         header->biRedMask |= (OPJ_UINT32)getc(IN) << 24;
 
+        if (!header->biRedMask) {
+            fprintf(stderr, "Error, invalid red mask value %d\n", header->biRedMask);
+            return OPJ_FALSE;
+        }
+
         header->biGreenMask  = (OPJ_UINT32)getc(IN);
         header->biGreenMask |= (OPJ_UINT32)getc(IN) << 8;
         header->biGreenMask |= (OPJ_UINT32)getc(IN) << 16;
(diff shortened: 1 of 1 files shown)


src/bin/jp2/convertbmp.c
```diff
@@ -831,6 +846,12 @@ opj_image_t* bmptoimage(const char *filename, opj_cparameters_t *parameters)
         bmpmask32toimage(pData, stride, image, 0x00FF0000U, 0x0000FF00U, 0x000000FFU,
                          0x00000000U);
     } else if (Info_h.biBitCount == 32 && Info_h.biCompression == 3) { /* bitmask */
+        if ((Info_h.biRedMask == 0U) && (Info_h.biGreenMask == 0U) &&
```
rouault: what happens if one or two of the component mask is 0 and not the 3 ? Wouldn't there be an issue too ?

rouault: OK got it. Actually given your checks at line 438, 448, 458, when we are at line 849 either the 3 values are at 0, or they are all non-zero.

Issue #1152, opened 2018-10-09
OPENJPEG null ptr dereference in openjpeg-2.3.0/src/bin/jp2/convert.c:2243


yngweijs:
# DESCRIPTION
OPENJPEG null ptr dereference in openjpeg-2.3.0/src/bin/jp2/convert.c:2243

VERSION
OPENJPEG-2.3.0

# GDB Output

#0  0x0000555555560a5d in imagetopnm (image=0x555555a22660, outfile=0x7fffffffbcac "out.ppm", force_split=0) at /home/pwnjs/Desktop/openjpeg/openjpeg-2.3.0/src/bin/jp2/convert.c:2243
#1  0x000055555555a0e1 in main (argc=5, argv=0x7fffffffddf8) at /home/pwnjs/Desktop/openjpeg/openjpeg-2.3.0/src/bin/jp2/opj_decompress.c:1692
#2  0x00007ffff71e7b97 in __libc_start_main (main=0x5555555592ed <main>, argc=5, argv=0x7fffffffddf8, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffdde8) at ../csu/libc-start.c:310
#3  0x00005555555567aa in _start ()

# GDB Information
pwndbg> print red
$1 = (int *) 0x0
pwndbg> print image->comps[compno].data
$2 = (OPJ_INT32 *) 0x0
pwndbg> c
Continuing.
Program received signal SIGSEGV, Segmentation fault.
0x0000555555560a5d in imagetopnm (image=0x555555a22660, outfile=0x7fffffffbcac "out.ppm", force_split=0) at /home/pwnjs/Desktop/openjpeg/openjpeg-2.3.0/src/bin/jp2/convert.c:2243
2243	                v = *red + adjustR;

# Analysis
image->comps[compno].data = NULL and it was assigned to red, so the program accesses to red, segment fault occurs.

# poc
Contact me if you need Poc file at yngweijs@gmail.com or mayfeelyang@gmail.com

yngweijs:
[291554_FlipDeter.zip](https://github.com/uclouvain/openjpeg/files/2458989/291554_FlipDeter.zip)

szukw000:
@rouault ,
2018-10-09, openjpeg: Latest commit 0e6a555 17 days ago

I have made one change in opj_decompress.c and one in convert.c .

[...]

PR #1156, opened 2018-10-18
Change in imagetopnm of convert.c in case image->comps[].data is NULL

szukw000:
See #1152 for the patch in opj_decompress: FlipDeter does not have an extension.
winfried

### DIFF
--- a/src/bin/jp2/opj_decompress.c
+++ b/src/bin/jp2/opj_decompress.c
@@ -538,12 +538,25 @@ static int infile_format(const char *fname)
         return ext_format;
     }
 
-    s = fname + strlen(fname) - 4;
+    s = fname + strlen(fname) - 1;
 
-    fputs("\n===========================================\n", stderr);
+#ifdef WIN32
+    while (s > fname && (*s != '.' && *s != '\\')) {
+        --s;
+    }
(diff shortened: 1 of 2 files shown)


src/bin/jp2/opj_decompress.c
```diff
@@ -538,12 +538,25 @@ static int infile_format(const char *fname)
         return ext_format;
     }
 
-    s = fname + strlen(fname) - 4;
+    s = fname + strlen(fname) - 1;
 
-    fputs("\n===========================================\n", stderr);
+#ifdef WIN32
```
rouault: it is legal to use forward slash on Windows too: c:/users/bla/...

szukw000:
@rouault ,
please see:

opj_decompress.c:
  line 449: const char* path_separator = "\\";

windirent.h:
static DIR *opendir(const char *dirname)
  line 357: strlen("\\*.*")

winfried

rouault:
> opj_decompress.c:
> line 449: const char* path_separator = "\";

That's not directly relevant. get_next_file() is free to use backslash to create a filename. But in get_file_format() , you receive a filename from the command line, and the user might have used forward slash.

PR #1159, opened 2018-10-21
Int overflow fixed

### LINKED ISSUE #1162: Int overflow in jp3d
When compressing a lot of slices (starting from 44 FullHD slices with 3 8bit components in our experiments) the rate values are high enough to cause an int overflow that leads to negative lengths and wrong results.

ichlubna:
When compressing a lot of slices (starting from 44 FullHD slices with 3 8bit components in our experiments) the rate values are high enough to cause an int overflow that leads to negative lengths and wrong results. The cast happens too late.
Fixes #1162

### DIFF
--- a/src/lib/openjp3d/tcd.c
+++ b/src/lib/openjp3d/tcd.c
@@ -226,10 +226,10 @@ void tcd_malloc_encode(opj_tcd_t *tcd, opj_volume_t * volume, opj_cp_t * cp,
         if (tcp->rates[j] <= 1) {
             tcp->rates[j] = 0;
         } else {
-            float num = (float)(tile->numcomps * (tile->x1 - tile->x0) *
-                                (tile->y1 - tile->y0) * (tile->z1 - tile->z0) * volume->comps[0].prec);
-            float den = (float)(8 * volume->comps[0].dx * volume->comps[0].dy *
-                                volume->comps[0].dz);
+            float num = (float)tile->numcomps * (tile->x1 - tile->x0) *
+                                (tile->y1 - tile->y0) * (tile->z1 - tile->z0) * volume->comps[0].prec;
+            float den = (float)8 * volume->comps[0].dx * volume->comps[0].dy *
+                                volume->comps[0].dz;
             den = tcp->rates[j] * den;
             tcp->rates[j] = (num + den - 1) / den;
         }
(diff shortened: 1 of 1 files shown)


src/lib/openjp3d/tcd.c
```diff
@@ -226,10 +226,10 @@ void tcd_malloc_encode(opj_tcd_t *tcd, opj_volume_t * volume, opj_cp_t * cp,
         if (tcp->rates[j] <= 1) {
             tcp->rates[j] = 0;
         } else {
-            float num = (float)(tile->numcomps * (tile->x1 - tile->x0) *
-                                (tile->y1 - tile->y0) * (tile->z1 - tile->z0) * volume->comps[0].prec);
-            float den = (float)(8 * volume->comps[0].dx * volume->comps[0].dy *
-                                volume->comps[0].dz);
+            float num = (float)tile->numcomps * (tile->x1 - tile->x0) *
+                                (tile->y1 - tile->y0) * (tile->z1 - tile->z0) * volume->comps[0].prec;
```
stweil: That avoids an overflow, but causes several float operations, so will be slower. Is this acceptable here?

Which part of the integer operations did overflow? Would it help to use larger integer types?

src/lib/openjp3d/tcd.c
```diff
@@ -226,10 +226,10 @@ void tcd_malloc_encode(opj_tcd_t *tcd, opj_volume_t * volume, opj_cp_t * cp,
...
+            float den = (float)8 * volume->comps[0].dx * volume->comps[0].dy *
```
stweil: Same comment and question as above. In any case the type cast is not necessary – writing `8.0f` is the same as `(float)8`.

src/lib/openjp3d/tcd.c
```diff
@@ -537,10 +537,10 @@ void tcd_init_encode(opj_tcd_t *tcd, opj_volume_t * volume, opj_cp_t * cp,
...
+                                volume->comps[0].dz;
```
stweil: See my previous comments.

src/lib/openjp3d/tcd.c
ichlubna:
```suggestion
            float den = 8.0f * volume->comps[0].dx * volume->comps[0].dy *
```
Looks probably better.

About the overflowing part, it depends on how big the slices are and how much of them we use. So it can occur basically at any of the multiplications.

Anyway casting the first operand to (long long) works as well but this expression is evaluated only once during compression and once during decompression so I don't think that the impact on the performance is that significant. When comparing the execution times the difference is not really visible.

xbarin02: @ichlubna  OpenJPEG should only require ANSI C (C89) compilers. The _long long_ is specified since the C99 version of the standard. Casting to _float_ is fine.

src/lib/openjp3d/tcd.c
```diff
@@ -537,10 +537,10 @@ void tcd_init_encode(opj_tcd_t *tcd, opj_volume_t * volume, opj_cp_t * cp,
         if (tcp->rates[j] <= 1) {
             tcp->rates[j] = 0;
         } else {
-            float num = (float)(tile->numcomps * (tile->x1 - tile->x0) *
-                                (tile->y1 - tile->y0) * (tile->z1 - tile->z0) * volume->comps[0].prec);
-            float den = (float)(8 * volume->comps[0].dx * volume->comps[0].dy *
-                                volume->comps[0].dz);
+            float num = (float)tile->numcomps * (tile->x1 - tile->x0) *
+                                (tile->y1 - tile->y0) * (tile->z1 - tile->z0) * volume->comps[0].prec;
+            float den = (float)8 * volume->comps[0].dx * volume->comps[0].dy *
```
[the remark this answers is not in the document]
ichlubna: Ah sorry forgot that one.

PR #1160, opened 2018-10-22
jp3d/jpwl convert: fix write stack buffer overflow

### LINKED ISSUE #1044: Similar vulnerable functions related to CVE-2017-14041
There are two functions similar to vulnerable function with id CVE-2017-14041 [http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-14041](url) with issue id (#997 ).

Below is the patch for CVE-2017-14041.
[https://github.com/uclouvain/openjpeg/commit/e5285319229a5d77bf316bb0d3a6cbd3cb8666d9](https://github.com/uclouvain/openjpeg/commit/e5285319229a5d77bf316bb0d3a6cbd3cb8666d9)
default

Below are two functions that similar to the vulnerable function.
1. src/bin/jp3d/convert.c (pgxtovolume)
```
    for (s = 0; s < sliceno; s++) {
        int pos = maxslice == sliceno ? s : pgxslicepos[s];
        f = fopen(pgxfiles[pos], "rb");
        if (!f) {
            fprintf(stdout, "[ERROR] Failed to open %s for reading !\n", pgxfiles[s]);
            return NULL;
        }    
        fprintf(stdout, "[INFO] Loading %s \n", pgxfiles[pos]);

        fseek(f, 0, SEEK_SET);
        fscanf(f, "PG%[ \t]%c%c%[ \t+-]%d%[ \t]%d%[ \t]%d", temp, &endian1, &endian2,
               signtmp, &prec, temp, &w, temp, &h); 

        i = 0; 
        sign = '+'; 
        while (signtmp[i] != '\0') {
            if (signtmp[i] == '-') {
                sign = '-'; 
            }    
            i++; 
        }
```
default

2. src/bin/jpwl/convert.c (pgxtoimage)

```
    fseek(f, 0, SEEK_SET);
    if (fscanf(f, "PG%[ \t]%c%c%[ \t+-]%d%[ \t]%d%[ \t]%d", temp, &endian1,
               &endian2, signtmp, &prec, temp, &w, temp, &h) != 9) { 
        fprintf(stderr,
                "ERROR: Failed to read the right number of element from the fscanf() function!\n");
        fclose(f);
        return NULL;
    }
```

default

I think there are vulnerabilities in those two functions, too. Therefore, there should be patches for those two functions.

hlef:
Missing buffer length formatter in fscanf call might lead to write stack buffer overflow.

(same patch as  e5285319229a5d77bf316bb0d3a6cbd3cb8666d9)

Fixes #1044 (CVE-2017-17480).

### DIFF
--- a/src/bin/jp2/convert.c
+++ b/src/bin/jp2/convert.c
@@ -2233,6 +2233,11 @@ int imagetopnm(opj_image_t * image, const char *outfile, int force_split)
                 opj_version(), wr, hr, max);
 
         red = image->comps[compno].data;
+        if (!red) {
+            fclose(fdest);
+            continue;
+        }
+
         adjustR =
             (image->comps[compno].sgnd ? 1 << (image->comps[compno].prec - 1) : 0);
 
--- a/src/bin/jp3d/convert.c
+++ b/src/bin/jp3d/convert.c
@@ -297,8 +297,8 @@ opj_volume_t* pgxtovolume(char *relpath, opj_cparameters_t *parameters)
         fprintf(stdout, "[INFO] Loading %s \n", pgxfiles[pos]);
 
         fseek(f, 0, SEEK_SET);
-        fscanf(f, "PG%[ \t]%c%c%[ \t+-]%d%[ \t]%d%[ \t]%d", temp, &endian1, &endian2,
-               signtmp, &prec, temp, &w, temp, &h);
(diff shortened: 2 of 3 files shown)


hlef:
FTR, I have added a second fix addressing #1152. A few comments:

* The first loop does not seem to be affected (line 2101), there are a few checks earlier during decoding forbidding the first tile to have zero size (not sure this is the expected behavior, but anyways). We might want to add a check there as well to be defensive, but it is not necessary to fix this issue.

* There might be other places in the code affected by this issue, but this patch does not address them.

PR #1163, opened 2018-10-31
Fix several memory and resource leaks

nforro:


src/bin/jp2/opj_compress.c
```diff
@@ -1739,6 +1739,8 @@ int main(int argc, char **argv)
     img_fol_t img_fol;
     dircnt_t *dirptr = NULL;
 
+    int ret;
```
rouault: it would be safer to initialize it to 0 here

nforro: Done.

PR #1165, opened 2018-11-02
Switch to GNUInstallDirs

a17r:
Provides standard install dirs without having to reinvent the wheel (and giving packagers standard overrideable install variables).


thesamesam:
Ping. I've just rebased these changes for 2.4.0 if there's any interest.

a17r:
> Ping - I've just rebased these changes for 2.4.0 if there's any interest?

Thanks for the reminder, rebased myself now. Anything different to your changes?

rouault:
I guess those changes have influences on the installation directories, particularly the include one, right ?

a17r:
Basically it replaces the homegrown vars with well established defaults. The default values do not change except for the cmake module location that is fixed.

See docs: https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

rouault:
I gave a try to the patch. . When enabling Doxygen, it installs the doc in /share/doc/OPENJPEG, whereas master installs in /share/doc/openjpeg-2.4. The uppercase in OPENJPEG is a bit annoying. I believe it would be better in lowercase (I don't think we care about the -2.4 suffix). Could that be changed ?

laumann:
Hi! I just rebased the same patch for 2.5.0 and will look into addressing your feedback @rouault

rouault:
superseded per https://github.com/uclouvain/openjpeg/pull/1424

PR #1168, opened 2018-11-23
Fix multiple potential vulnerabilities and bugs

YangY-Xiao:
Fix multiple potential vulnerabilities and bugs

### DIFF
--- a/src/bin/jpwl/convert.c
+++ b/src/bin/jpwl/convert.c
@@ -99,15 +99,10 @@ struct tga_header {
 };
 #endif /* INFORMATION_ONLY */
 
-static unsigned short get_ushort(unsigned short val)
+/* Returns a ushort from a little-endian serialized value */
+static unsigned short get_tga_ushort(const unsigned char *data)
 {
-
-#ifdef OPJ_BIG_ENDIAN
-    return (((val & 0xff) << 8) + (val >> 8));
-#else
-    return (val);
-#endif
-
+    return data[0] | (data[1] << 8);
 }
 
 #define TGA_HEADER_SIZE 18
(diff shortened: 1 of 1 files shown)


src/bin/jpwl/convert.c
```diff
@@ -135,15 +131,19 @@ static int tga_readheader(FILE *fp, unsigned int *bits_per_pixel,
     id_len = (unsigned char)tga[0];
     cmap_type = (unsigned char)tga[1];
     image_type = (unsigned char)tga[2];
-    cmap_index = get_ushort(*(unsigned short*)(&tga[3]));
+    cmap_index = get_tga_ushort(*(unsigned short*)(&tga[3]));
     cmap_len = get_ushort(*(unsigned short*)(&tga[5]));
+    cmap_len = get_tga_ushort(*(unsigned short*)(&tga[5]));
```
rouault: I suspect this doesn't compile... Line 135 should be removed as it is effectively overrideen by line 136. Similar instances below

Young-X: I have fixed that by make a new pr.

Young-X:
Ping

ret2libc:
> @Young-X I had to revert that commit in e1740e7 since it didn't compile. I'm disappointed you submit commits that don't even compile...

@rouault it seems commit https://github.com/uclouvain/openjpeg/commit/e1740e7ce79d0a1676db4da0f4189b64e85f52cb was supposed to fix CVE-2018-20846, which is about out-of-bound accesses in pi_next_lrcp, pi_next_rlcp, pi_next_rpcl, pi_next_pcrl, pi_next_rpcl, and pi_next_cprl . With that commit reverted, is there a patch for that CVE? Thanks.

rouault:
> is there a patch for that CVE?

no

PR #1171, opened 2018-11-30
gcc-8.2.0 warnings: possible buffer overflow

szukw000:
gcc-8.2.0 warnings: possible buffer overflow using strcpy()


szukw000:
@rouault ,
I do not understand this:

Could not find executable C:/projects/openjpeg/build/bin/testempty2.exe

EXE on a LINUX machine?

find openjpeg -name "testempty2*"
openjpeg/tests/unit/testempty2.c

winfried

PR #1176, opened 2019-01-09
Add parsing of pxfm boxes

DerouineauNicolas:
This PR adds the parsing of pxfm boxes defined in  15444-2:2004/Amd.3:2015.

I'm not sure about the design of the opj_jp2_pixel_format_values_t. I have been thinking about using an Union instead of a Struct, as the standard is only defining either mantissa, exponent or number of fractionnal bits.

### DIFF
--- a/src/lib/openjp2/jp2.c
+++ b/src/lib/openjp2/jp2.c
@@ -1493,34 +1493,38 @@ static OPJ_BOOL opj_jp2_read_pxfm(opj_jp2_t * jp2,
     (void)p_pxfm_header_size;
 
     opj_read_bytes(p_pxfm_header_data, &l_value,
-                       2);
+                   2);
 
     num_channel = (OPJ_UINT16) l_value;
 
-    p_pxfm_header_data+=2;
+    p_pxfm_header_data += 2;
 
     if (jp2->numcomps != num_channel) {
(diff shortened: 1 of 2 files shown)


src/lib/openjp2/jp2.c
```diff
@@ -1455,6 +1474,63 @@ static OPJ_BOOL opj_jp2_read_cdef(opj_jp2_t * jp2,
...
+        p_pxfm_header_data += 2;
+        Channel_index = (OPJ_UINT16) l_value;
+        opj_read_bytes(p_pxfm_header_data, &l_value,
+                       2);
+        p_pxfm_header_data += 2;
+        if (Channel_index < jp2->numcomps) {
+            jp2->pixel_format[Channel_index].pixel_format_type = ((OPJ_UINT16) l_value) &
+                    0xF000;
+
+            if (jp2->pixel_format[Channel_index].pixel_format_type == 0 ||
```
rouault: given the above masking, it is not possible to have pixel_format_type being 0 or 4...

src/lib/openjp2/jp2.h
```diff
@@ -143,6 +144,31 @@ typedef struct opj_jp2_comps {
     OPJ_UINT32 bpcc;
 } opj_jp2_comps_t;
 
+
+typedef enum {
```
rouault: this enumeration type appears to be unused ?

PR #1185, opened 2019-03-16
Fix several potential vulnerabilities 

Young-X:


rouault:
Can you fix the style issue as raised in https://travis-ci.org/uclouvain/openjpeg/jobs/507178238 ?
See tips at the end of the log:
```
Tips to prevent and resolve:
* Enable WITH_ASTYLE in your cmake configuration to format C++ code
* Install autopep8 (>= 1.2.1) to format python code
* Use "scripts/astyle.sh file" to fix the now badly indented files
* Consider using scripts/prepare-commit.sh as pre-commit hook to avoid this
  in the future (ln -s scripts/prepare-commit.sh .git/hooks/pre-commit) or
  run it manually before each commit.
```

PR #1190, opened 2019-04-24
CMakeLists.txt: remove some evil codes

mtdcy:
when building as sub directory in project, these evil codes will change other project target's output


rouault:
This PR breaks all builds. Closing

Issue #1207, opened 2019-08-01
Decode fails


acdha:
While troubleshooting an odd image display on https://chroniclingamerica.loc.gov/lccn/sn83025887/1791-12-22/ed-1/seq-4/ I noticed some inconsistencies between codecs handling the file, available directly as https://chroniclingamerica.loc.gov/lccn/sn83025887/1791-12-22/ed-1/seq-4.jp2

* The MacOS CoreImage framework fails to decode it entirely (IIRC this at least used to be a licensed, older version of Kakadu)
* Kakadu v7.10.2 decodes it correctly
* OpenJPEG v2.3.1 generates an image of the expected dimensions but it has the image data displayed three times side-by-side across the width of the image:
![0004](https://user-images.githubusercontent.com/46565/62304614-5189a480-b44c-11e9-9712-5b0d9a6bbf8f.jpg)

If I take the same image and run it through kdu_transcode, it produces the expected image:

```sh
$ kdu_transcode -i 0004.jp2 -o 0004-transcode.j2k
$ opj_decompress -i 0004-transcode.j2k -o 0004-transcode.png
```

szukw000:
@acdha ,
the National Gazette shows three pages, The 'Text | PDF | JP2 (2.5 MB)' JP2 file
shows 3 pages with OPENJPEG.
Why should the National Gazette be wrong and KDU be right?
winfried

acdha:
@szukw000 That issue had 4 pages, not three, and note that it's displaying the same page 3 times rather than 3 separate pages (here's what an unaffected batch looks like: https://chroniclingamerica.loc.gov/lccn/sn91066782/1919-08-05/ed-1/seq-1/). 

This was reported to me as a bug but it's specific to certain batches which apparently slipped through quality-review because the 4 JP2 files in that issue all decode to the expected single page each using Photoshop, Kakadu, and Aware.

szukw000:
@acdha ,
the file 0004.jp2 has an ICC profile. Excluding to use this ICC profile:
```
OPJ_DECOMPRESS:1609: icc_profile_len(560)
```
I got a one image result. Must look next day what is going wrong.
winfried

acdha:
If memory serves, we ran into this with an older Kakadu version as well. I’ll check the history tomorrow.

acdha:
[...]

PR #1209, opened 2019-08-12
ICC data in color_apply_icc_profile() collide with number of CDEF channels

szukw000:
See issue #1207: files 0001.jp2, 0003.jp2 and 0004.jp2 have 1 component
(GRAY) but their ICC data need > 2 components.


szukw000:
Closes #1209

Issue #1210, opened 2019-09-03
Bitstream parser lacks validation for some segment marker parameter values...


sebras:
When manually analyzing the bit stream from issue #1158 it was evident that the underlying issue is that `SGcod(C)`, `SPcod(H)` and `SPcoc(E)` were out of range.

This is was never detected by OpenJPEG because `opj_j2k_read_cod()` and `opj_j2k_read_SPCod_SPCoc()` never validate these values after reading them. Other parameter values from the same segment markers are validated however.

I will attempt to provide a sane pull request for addressing this.

PR #1211, opened 2019-09-04
Add check to validate SGcod/SPcoc/SPcod parameter values.

sebras:
This is my attempt to fix issue #1210 

Merging this commit may need changes in `openjpeg-data` as well for the test files since three files now fail consistently, but the `baseline` directory in `openjpeg-data` contains fully parsed header information.

If you prefer I could change the code to only warn in these circumstances (and update the test cases accordingly), but it that would a step away from the specification. Let me know what you think.

### DIFF
--- a/src/lib/openjp2/j2k.c
+++ b/src/lib/openjp2/j2k.c
@@ -9739,9 +9739,9 @@ static OPJ_BOOL opj_j2k_read_SPCod_SPCoc(opj_j2k_t *p_j2k,
         return OPJ_FALSE;
     }
 
-    opj_read_bytes(l_current_ptr, &l_tccp->numresolutions,
-                   1);              /* SPcox (D) */
-    ++l_tccp->numresolutions;                                                                               /* tccp->numresolutions = read() + 1 */
+    /* SPcod (D) / SPcoc (A) */
+    opj_read_bytes(l_current_ptr, &l_tccp->numresolutions, 1);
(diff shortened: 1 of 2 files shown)


sebras:
*sigh* I had forgotten about the code formatting check run by CI... The log did mention that it can be run locally which was really helpful! :)

What is problematic is that the ABI check fails, I can't run it locally and it log doesn't explain how I may have changed the ABI. Do you mind helping out with that? Perhaps it was just a momentary anomaly.

rouault:
> What is problematic is that the ABI check fails

This isn't related to your change, and also fails with master.

sebras:
> This isn't related to your change, and also fails with master.

Oh, that's a relief at least. :) Are you otherwise happy with the pull request?

rouault:
> Are you otherwise happy with the pull request?

My brain has not been involved with JPEG2000 crazyness for a few months now (might be soon perhaps), so I'm not sure I've an informed opinion right now. I'm not sure to understand if those checks are nice to have, or if they fix crashes later in the code
I believe mct > 1 might be valid in an extension profile of JPEG2000
It is a bit unfortunate existing test cases now rejected since they tested other parts of the code. Wondering if they couldn't be manually patched to not be rejected by the new checks ?

rouault:
ABI check issue is solved.

rouault:
OK, on review & testing with Kakadu, those additional checks are reasonable. Merging

PR #1212, opened 2019-09-09
Buffer-based streams.

nikkoara:
This existed previously as #583 and stayed here for a number of years. If there is still interest in it, here it is, in one commit.

I left the management of the buffer memory to the user instead of hooking it up in the logic of opj_stream_destroy. Make sure you dispose of the memory the library allocates during compression.


rouault:
I don't think the opj_buffer_info structure should be exposed in openjpeg.h, in particular because the use of the cur pointer is purely an implementation detail of openjpeg. And for the sake of const correctness, we should probably distinguish the input and output cases

So the API should rather be 
```
OPJ_API opj_stream_t* OPJ_CALLCONV
opj_stream_create_input_stream_from_buffer (const void*, size_t);

OPJ_API opj_stream_t* OPJ_CALLCONV
opj_stream_create_output_stream_from_buffer (void*, size_t);
```
Some testing of this implementation might be welcome as well.
The CI also complains about code formatting

nickdademo:
Any plan to get this PR merged into master?

Issue #11, opened 2009-11-06
OpenJPEG cannot decode this image


gcode-importer:
Originally reported on Google Code with ID 11

```

The attached JP2 image is a color image with YCbCr components, the Cb and
Cr components subsampled. The image renders fine with kdu_render, but
j2k_to_image seems to get stuck in an infinite loop.

What steps will reproduce the problem?
1. j2k_to_image -i reference.jp2 -o foo.ppm

What is the expected output? What do you see instead?
I've attached a .png version of the output that I get from kdu_render.

What version of the product are you using? On what operating system?
OpenJPEG V2 Alpha, on Linux/x86_64 (SuSE 10.3)

Please provide any additional information below.


```

Reported by [steveicarus](https://github.com/steveicarus) on 2009-11-06 17:09:36

<hr>
- _Attachment: [reference-1024x1024x3-4tiles-YCbCr-4:2:0.jp2](https://storage.googleapis.com/google-code-attachments/openjpeg/issue-11/comment-0/reference-1024x1024x3-4tiles-YCbCr-4:2:0.jp2)_
- _Attachment: reference-1024x1024x3-4tiles-YCbCr-4:2:0.png<br>![reference-1024x1024x3-4tiles-YCbCr-4:2:0.png](https://storage.googleapis.com/google-code-attachments/openjpeg/issue-11/comment-0/reference-1024x1024x3-4tiles-YCbCr-4:2:0.png)_

PR #1223, opened 2019-12-12
Ctest development for Travis test #11

CBreshears:
Added known failures file to match current Travis OS version and compiler used. This includes tests named 
- NR-DEC-issue205.jp2-43-decode-md5
- NR-DEC-issue208.jp2-69-decode-md5

These tests are still failing in the most recent version of OpenJPEG library and gcc 5.4.0.


rouault:
This pull request mixes many different things. Please rework it to just contain the strict minimum

PR #1230, opened 2020-01-13
tests/fuzzers: link fuzz binaries using $LIB_FUZZING_ENGINE.

Dor1s:
This was changed some time ago (https://google.github.io/oss-fuzz/getting-started/new-project-guide/) but the build didn't fail as there is a fallback mechanism. The main advantage of the new approach is that for libFuzzer this produces more performant binaries (as `$LIB_FUZZING_ENGINE` expands into `-fsanitize=fuzzer`, which links libFuzzer from the compiler-rt, allowing better optimization tricks).

I'm also experimenting with dataflow (https://github.com/google/oss-fuzz/issues/1632) on your project, and the dataflow config doesn't have a fallback (as it's a new configuration), therefore I'm proposing a change to migrate from `-lFuzzingEngine` to `$LIB_FUZZING_ENGINE`.

### DIFF
--- a/tests/fuzzers/GNUmakefile
+++ b/tests/fuzzers/GNUmakefile
@@ -8,5 +8,6 @@ fuzzingengine.o: fuzzingengine.c
 
 dummyfuzzers: fuzzingengine.o
 	$(AR) r libFuzzingEngine.a fuzzingengine.o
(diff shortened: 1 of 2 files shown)


rouault:
tests/fuzzers/GNUmakefile should be modified to define LIB_FUZZING_ENGINE=-lFuzzingEngine

Dor1s:
Thanks @rouault, added, let's see if it goes through

Dor1s:
Travis is happy!

PR #1242, opened 2020-04-06
Build meson support

dabrain34:


rouault:
I'm not sure we want to add a second build system into the library...

dabrain34:
Sorry I realized that I did not add much to my PR description.

My idea with this PR was to provide the possibility to use openjpeg as a subproject of another meson project (gst-build). 
I can understand that you dont want to maintain multiple buildsystem.

But as you can see, I did not touch anything in CMake and we can have also the power and flexiility of the trendy Meson.

rouault:
openjpeg is insufficiently maintained (by me in particular :-)), so adding more stuff that could rot easily (as I won't use it personnally) is not super compelling.

You might want to raise the issue on the openjpeg mailing list and see if it raises interest from other stakeholders.

by the way the change in libtiff should likely be reverted. libtiff has an upstream (https://gitlab.com/libtiff/libtiff), and openjpeg should not modify it, unless resynchronizing with that upstream

dabrain34:
No problem I will drop the `indent` commit in libtiff.

I understand your point regarding the support of 2 build system :)

 I just wanted to let you know first of the existence of this port.

PR #1251, opened 2020-05-19
added support for partial bitstream decoding

chafey:
This PR adds support for partial bitstream decoding.  Given a partial bitstream, it will still reconstruct an image with the data it has available.


ValZapod:
Also see implementation https://github.com/jarrelscy/openjpeg/commits/master

rouault:
I've no idea what the last commit "added devcontainer and build.sh"  does, and it should be removed from this PR

Issue #1259, opened 2020-06-22
Memory leak when failing to allocate object...


sebras:
I found a memory leak by using a custom allocator where I can control when it returns NULL.

This is the relevant parts of two backtraces showing the locations of the leaks:

```
    #0 0x555ea46e8223 in opj_calloc
    #1 0x555ea4d44024 in opj_image_create0 openjpeg/src/lib/openjp2/image.c:36
    #2 0x555ea4f91773 in opj_j2k_read_header openjpeg/src/lib/openjp2/j2k.c:7535
    #3 0x555ea4d46f43 in opj_read_header openjpeg/src/lib/openjp2/openjpeg.c:432

    #0 0x555ea46e81c6 in opj_malloc
    #1 0x555ea4d455b3 in opj_copy_image_header openjpeg/src/lib/openjp2/image.c:192
    #2 0x555ea4f91814 in opj_j2k_read_header openjpeg/src/lib/openjp2/j2k.c:7541
    #3 0x555ea4d46f43 in opj_read_header openjpeg/src/lib/openjp2/openjpeg.c:432
```

This is the backtrace of where the allocator returns `NULL`, causing the two leaks above:
```
#0  0x0000555555833224 in opj_calloc (n=100, size=24)
#1  0x00005555560f1a70 in opj_j2k_allocate_tile_element_cstr_index (p_j2k=0x613000001548) at openjpeg/src/lib/openjp2/j2k.c:10658
#2  0x00005555560dc81e in opj_j2k_read_header (p_stream=0x60c000008808, p_j2k=0x613000001548, p_image=0x7fffffff9d70, p_manager=0x6100000005a8) at openjpeg/src/lib/openjp2/j2k.c:7544
#3  0x0000555555e91f44 in opj_read_header (p_stream=0x60c000008808, p_codec=0x610000000548, p_image=0x7fffffff9d70) at openjpeg/src/lib/openjp2/openjpeg.c:432
```

Looking at the [source code](https://github.com/uclouvain/openjpeg/blob/master/src/lib/openjp2/j2k.c#L8286) it is obvious that `p_image` is never freed if `opj_j2k_allocate_tile_element_cstr_index()` returns an error.

ValZapod:
Can be closed.

PR #1260, opened 2020-06-22
openjp2: Plug image leak when failing to allocate codestream index.

sebras:
This fixes issue #1259.


sebras:
Thank you for being extremely responsive in merging this set of leak and crash fixes. It's sincerely appreciated. :)

PR #1262, opened 2020-06-28
opj_decompress: fix double-free on input directory with mix of valid and invalid images

### LINKED ISSUE #1261: Heap use-after-free
There is a heap use-after-free vulnerability in openjpeg/build/bin/libopenjp2.so.7. 
I would also like to request a CVE if that's okay or I can do it myself if you're busy, no worries. 

Here is the ASAN output:
[out.txt](https://github.com/uclouvain/openjpeg/files/4842045/out.txt)

Here is the command executed to reproduce the issue: 
./opj_decompress -ImgDir Inputs/ -OutFor PGM

If you would also like the files I have in the Inputs/, let me know and I can find a way to provide them.

rouault:
Fixes #1261

Credits to @Ruia-ruia for reporting and analysis.

### DIFF
--- a/src/bin/jp2/opj_decompress.c
+++ b/src/bin/jp2/opj_decompress.c
@@ -1316,10 +1316,6 @@ static opj_image_t* upsample_image_components(opj_image_t* original)
 int main(int argc, char **argv)
 {
     opj_decompress_parameters parameters;           /* decompression parameters */
-    opj_image_t* image = NULL;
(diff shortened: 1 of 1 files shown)


rouault:
@Ruia-ruia are you requesting a CVE number ? If so, provide it to me so I can tag the commit message with it. If not, I'll just merge that as it

PR #1290, opened 2020-11-30
Set parameters->irreversible = 0 to avoid buffer-overflow

zodf0055980:
When using `-IMF`, it set `parameters->irreversible = 1` in j2k:6984
```c
if (profile == OPJ_PROFILE_IMF_2K ||
            profile == OPJ_PROFILE_IMF_4K ||
            profile == OPJ_PROFILE_IMF_8K) {
        /* 9-7 transform */
        parameters->irreversible = 1;
    }
```
However, when `opj_j2k_is_imf_compliant` return `OPJ_FALSE` to use Non-IMF codestream, it doesn't set `parameters->irreversible` to zero in j2k:7732
```c
        if (!opj_j2k_is_imf_compliant(parameters, image, p_manager)) {
            parameters->rsiz = OPJ_PROFILE_NONE;
            // parameters->irreversible = 0;  <- fix
        }
```
So in `opj_dwt_calc_explicit_stepsizes` in dwt.c:1963, it does IMF code to cause buffer-over-flow.
```c
        if (tccp->qntsty == J2K_CCP_QNTSTY_NOQNT) {
            stepsize = 1.0;
        } else {
            OPJ_FLOAT64 norm = opj_dwt_norms_real[orient][level];
            stepsize = (1 << (gain)) / norm;
        }
```

### DIFF
--- a/src/lib/openjp2/j2k.c
+++ b/src/lib/openjp2/j2k.c
@@ -7730,6 +7730,7 @@ OPJ_BOOL opj_j2k_setup_encoder(opj_j2k_t *p_j2k,
         opj_j2k_set_imf_parameters(parameters, image, p_manager);
         if (!opj_j2k_is_imf_compliant(parameters, image, p_manager)) {
             parameters->rsiz = OPJ_PROFILE_NONE;
+            parameters->irreversible = 0;
         }
     } else if (OPJ_IS_PART2(parameters->rsiz)) {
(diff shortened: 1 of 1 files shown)


rouault:
I don't believe this fix is correct. The issue you try to fix can probably happen in non-IMF mode

rouault:
superseded by https://github.com/uclouvain/openjpeg/pull/1292

PR #1306, opened 2020-12-09
Use INC_DIR for OPENJPEG_INCLUDE_DIRS (fixes uclouvain#1174)

matthew-sharp:
This fixes the case where the cmake file is accessed via a symlink.

I'm not sure I understand the original code and why all the complexity is needed. If we do need all that, then we may need to canonicalize SELF_DIR.


Jeinzi:
Thank you! :)

PR #1307, opened 2020-12-14
Set ${OPENJPEG_INSTALL_DOC_DIR} to DESTINATION of HTMLs

lemniscati:
Use `${OPENJPEG_INSTALL_DOC_DIR}` as `DESTINATION` of HTML documents, instead of `share/doc`.


lemniscati:
@rouault Thank you.

PR #1310, opened 2020-12-21
Change defined WIN32

Jamaika1:


rouault:
What error does that fix ?

Jamaika1:
I don't want any warning messages for OPJ_STATIC on windows in GCC.
```
warning: visibility attribute not supported in this configuration; ignored [-Wattributes]
```

PR #1315, opened 2020-12-29
Import files tiff and yuv(raw)

Jamaika1:


rouault:
There are 3 pull requests with the same title. Could you close the one(s) that are no longer relevant ?

Jamaika1:
Three different files. All corrections are important.

PR #1321, opened 2021-01-07
Fix cmake file with DESTDIR

ffontaine:
This reverts commit 6558637 which breaks cross-compilation of poppler under buildroot (because of `DESTDIR` usage). Instead, `ABSOLUTE` is replaced by `REALPATH` to resolve symlinks (https://cmake.org/cmake/help/latest/command/get_filename_component.html)


rouault:
@matthew-sharp Can you test this PR ?

matthew-sharp:
This PR works for me. Apologies for breaking cross-compile.

rouault:
> This PR works for me

Thanks for the feedback

PR #1345, opened 2021-04-26
emscripten: disable big endian test

msabwat:
Hi, 

I think this test should not run on the emscripten platform because of :
https://github.com/emscripten-core/emscripten/blob/dff33368427fba16745c8ce52f11484a67b2855d/cmake/Modules/TestBigEndian.cmake#L5

changing CMAKE_ROOT to point to `..,/emsdk/upstream/emscripten/cmake` does not solve the issue, as there are some .cmake files that are needed and are not on emsdk.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,8 +162,10 @@ endif()
 
 #-----------------------------------------------------------------------------
 # Big endian test:
+if (!EMSCRIPTEN)
 include (${CMAKE_ROOT}/Modules/TestBigEndian.cmake)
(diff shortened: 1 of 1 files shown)


rouault:
Could you rebase your pull request on top of latest master (which has fixes for the CI failures) ?

msabwat:
Hi @rouault, thank you. Done.

attilaolah:
For the record, I think the solution here is not to change `CMAKE_ROOT`, but to instead use `${CMAKE_MODULE_PATH}` instead of `${CMAKE_ROOT}/Modules`. Emscripten sets this variable to the path containing the correct endianness checks.

PR #1339, opened 2021-03-24
openmj2/opj_includes.h: Do not define 'lrintf' on MSVC since 16.9.2

emmenlau:
This PR adds a minor fix for building with Visual Studio 16.9 ore newer. The newer versions have a builtin lrintf, so the current code was giving error:
```
[...]\src\lib\openmj2\opj_includes.h(97): error C2169: 'lrintf': intrinsic function, cannot be defined
```

This PR closes #1333

PR #1350, opened 2021-05-04
Remove obsolete components JPWL, JP3D and MJ2

rouault:


nist-anka:
#1339 still happens when building the release version of an app in Visual Studio 2019 v. 16.11 
debug version builds fine

PR #1354, opened 2021-06-01
fix #1345: don't remove big endian test for other platforms

msabwat:
thank you to  @ePirat for finding the issue, it turns out that !EMSCRIPTEN does not evaluate to 1 on other platforms, so the the #1345 actually disabled the test for all platforms.

sorry about the churn


msabwat:
#1355

PR #1366, opened 2021-07-12
opj_j2k_is_imf_compliant: Fix out of bounds access

sebras:
This was discovered when scanning the code using Coverity.


rouault:
I'd prefer if you'd put the /* Validate sublevel */ logic as a else clause of the ``if (mainlevel > OPJ_IMF_MAINLEVEL_MAX)``. That should fix the out of bound access, while still allowing more warnings to be emitted

sebras:
@rouault Ok, that's easy enough. Done. :)

PR #1377, opened 2021-09-12
Change integer for version libtiff 4.3.0

Jamaika1:


rouault (changes requested):
Could you please change that to use unconditionally the standard types (uint16_t / uint32_t) ? They are safe to use with older libtiff versions

PR #1381, opened 2021-09-25
Added support for high throughput (HTJ2K) decoding.

rouault:
There are a few limitations:
- mixed mode (HT and regular code blocks) is not supported
- ROI in HT blocks is not supported
- Placeholder passes are not supported.
- MultiHT sets are not support, only a singleHT set.
- there are known issues with some compliance testing files related to
  the parsing of packet header

Continuation of https://github.com/uclouvain/openjpeg/pull/1374, which is a work by @aous72


rouault:
@aous72 Does the PR text / commit message captures well the (known) limitations of the implementation ?

aous72:
@rouault 
Thank you for merging this.

> Does the PR text / commit message captures well the (known) limitations of the implementation ?

All the points are correct.  Perhaps, I would replace "only up to 3 coding passes are supported" with:
- Placeholder passes are not supported.
- MultiHT sets are not support, only a singleHT set.

The singleHT set can have up to 3 passes.  Feel free to modify the text.

Thank you.

aous72:
@rouault 
So, if I want to do more development, what should I do? clone it again?

rouault:
> Do you have an idea how to test it big endian?

Travis-CI has a s390x big endian target but given their new economic model, it is impractical for open source projects if you don't subscribe to their service. Otherwise if you don't have access to a big endian box, you can potentially use qemu full system emulation with a big endian emulated machine (will be rather slow).  Honestly, I wouldn't put too much effort on that given the very very low market shared of big endian nowadays. Efforts on improving compliance and speed would be much better time investment.

> A second question: If I want to do more development, what should I do? clone it again?

no, normally, you would just do
```
git checkout master
git pull origin
git checkout -b new_branch
```

but as your working branch was master (not recomended practice), you perhaps need to do something along
```
git checkout master
git fetch origin master
git reset --hard origin/master  # beware: will loose all your local changes !
git checkout -b new_branch
```
I'm assuming your ``origin`` remote points to uclouvain/openjpeg

aous72:
Thank you for your detailed explanation.

malaterre:
@aous72 

> Do you have an idea how to test it big endian?

You can request access to a POWER machine here:

* https://openpower.ic.unicamp.br/minicloud/

As mentionned by @rouault this should not be a top priority anyway.

aous72:
You can request access to a POWER machine here:
    https://openpower.ic.unicamp.br/minicloud/

@malaterre Very useful suggestion.  Thank you.

PR #1388, opened 2021-11-15
Allow configuration of the number of guard bits when encoding

cth103:
The recently-released SMPTE DCP Bv2.1 Application Profile (link below) says that the number of guard bits in the `QCD` marker shall be 1 for 2K content and 2 for 4K content.  This change allows the number of guard bits to be configured, so that users of openjpeg have the control they need to meet the specification.

https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=9161348


rouault:
Superseded per https://github.com/uclouvain/openjpeg/pull/1403

PR #1390, opened 2021-12-05
Fix some typos (found by codespell)

stweil:

### DIFF
--- a/src/lib/openjp2/j2k.c
+++ b/src/lib/openjp2/j2k.c
@@ -8829,7 +8829,7 @@ static OPJ_BOOL opj_j2k_read_header_procedure(opj_j2k_t *p_j2k,
         if (l_marker_handler->id == J2K_MS_UNK) {
             if (! opj_j2k_read_unk(p_j2k, p_stream, &l_current_marker, p_manager)) {
                 opj_event_msg(p_manager, EVT_ERROR,
-                              "Unknown marker have been detected and generated error.\n");
+                              "Unknown marker has been detected and generated error.\n");
                 return OPJ_FALSE;
             }
 
--- a/src/lib/openjp2/j2k.c
+++ b/src/lib/openjp2/j2k.c
@@ -50,7 +50,7 @@
 /*@{*/
 
 /**
- * Sets up the procedures to do on reading header. Developpers wanting to extend the library can add their own reading procedures.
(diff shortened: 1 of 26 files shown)


src/lib/openjp2/j2k.c
```diff
@@ -8829,7 +8829,7 @@ static OPJ_BOOL opj_j2k_read_header_procedure(opj_j2k_t *p_j2k,
         if (l_marker_handler->id == J2K_MS_UNK) {
             if (! opj_j2k_read_unk(p_j2k, p_stream, &l_current_marker, p_manager)) {
                 opj_event_msg(p_manager, EVT_ERROR,
-                              "Unknow marker have been detected and generated error.\n");
+                              "Unknown marker have been detected and generated error.\n");
```
stweil, line 8832: This looks also like a singular / plural mismatch. Should it be `markers have` or `marker has`?

rouault: should be singular

rouault:
the astyle check fails because you've touched files that weren't compliant with it before. Would you mind running astyle on them first to have a reindent-only commit, and then fix the typos in a separate commit ?

stweil:
Would it be also okay to add the astyle commit after the typo fixes (which required some manual work)?

rouault:
> Would it be also okay to add the astyle commit after the typo fixes (which required some manual work)?

yeah, that can do it too

src/lib/openjp2/j2k.c
stweil: Fixed in 2nd commit.

stweil:
Only a single source file had to be re-formatted (done in 3rd commit).

src/lib/openjp2/j2k.c
```diff
@@ -50,7 +50,7 @@
 /*@{*/
 
 /**
- * Sets up the procedures to do on reading header. Developpers wanting to extend the library can add their own reading procedures.
+ * Sets up the procedures to do on reading header. Developers wanting to extend the library can add their own reading procedures.
```
stweil, line 53: @rouault, you could use the same change in your GitHub profile.

rouault: thanks :-) Frenchism corrected

PR #1396, opened 2022-01-13
Added overflow check for CVE-2021-29338 

Eharve14:
Added a multiplication check before the call to calloc, see my comment on 79c7d7a

### DIFF
--- a/src/bin/jp2/opj_compress.c
+++ b/src/bin/jp2/opj_compress.c
@@ -1957,6 +1957,11 @@ int main(int argc, char **argv)
     /* Read directory if necessary */
     if (img_fol.set_imgdir == 1) {
         num_images = get_num_images(img_fol.imgdirpath);
+        if((num_images > SIZE_MAX/(OPJ_PATH_LEN * sizeof(char))){
+            fprintf(stdout, "Max images exceeded\n");
+            ret = 0;
+            goto fin;
+        } else {
         dirptr = (dircnt_t*)malloc(sizeof(dircnt_t));
         if (dirptr) {
             dirptr->filename_buf = (char*)calloc(num_images, OPJ_PATH_LEN * sizeof(
@@ -1973,6 +1978,7 @@ int main(int argc, char **argv)
         if (load_images(dirptr, img_fol.imgdirpath) == 1) {
             ret = 0;
             goto fin;
+        }
         }
         if (num_images == 0) {
             fprintf(stdout, "Folder is empty\n");
--- a/src/bin/jp2/opj_compress.c
+++ b/src/bin/jp2/opj_compress.c
@@ -1957,11 +1957,6 @@ int main(int argc, char **argv)
     /* Read directory if necessary */
     if (img_fol.set_imgdir == 1) {
         num_images = get_num_images(img_fol.imgdirpath);
-        if((num_images > SIZE_MAX/(OPJ_PATH_LEN * sizeof(char)))){
-            fprintf(stdout, "Max images exceeded\n");
-            ret = 0;
-            goto fin;
-        } else {
         dirptr = (dircnt_t*)malloc(sizeof(dircnt_t));
         if (dirptr) {
             dirptr->filename_buf = (char*)calloc(num_images, OPJ_PATH_LEN * sizeof(
@@ -1971,14 +1966,13 @@ int main(int argc, char **argv)
                 ret = 0;
                 goto fin;
             }
-            for (i = 0; i < num_images; i++) {
+            for (size_t i = 0; i < num_images; i++) {
                 dirptr->filename[i] = dirptr->filename_buf + i * OPJ_PATH_LEN;
             }
         }
         if (load_images(dirptr, img_fol.imgdirpath) == 1) {
             ret = 0;
             goto fin;
-        }
         }
         if (num_images == 0) {
             fprintf(stdout, "Folder is empty\n");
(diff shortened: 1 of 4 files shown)


rouault:
> Added a multiplication check before the call to calloc, see my comment on [79c7d7a](https://github.com/uclouvain/openjpeg/commit/79c7d7af598b778c3cdcb455df23d50efc95eb3c)

I saw your comment, but calloc() should really check for the overflow. I'd expect only old and unmaintained platform to have a buggy implementation and we probably don't care about them. Or said otherwise: can you quote one recent enough platform of interest whose calloc() wouldn't check the overflow of the multiplication of the argument ?

rouault:
ok, answering myself I see the list is in the link : https://www.avertium.com/blog/overview-badalloc-vulnerabilities.
Well none of them qualify as of major players as far as I'm concerned to justify cluttering openjpeg codebase. This is a bug of those C libraries, not openjpeg.

Eharve14:
Rodger that, just wanted to bring some attention to the topic since I had seen the update and I had done some research on the topic.

Eharve14:
Specifically when the strcpy function is called in the load_images function.

src/bin/jp2/opj_compress.c
```diff
@@ -1966,7 +1966,7 @@ int main(int argc, char **argv)
                 ret = 0;
                 goto fin;
             }
-            for (i = 0; i < num_images; i++) {
+            for (size_t i = 0; i < num_images; i++) {
```
rouault, line 1969: we have still an old CI platform (VS2010 !) that doesn't support C99. Could you just add the cast to size_t in the multiplications ?

Eharve14: If you do a cast after the multiplication the multiplication will be handled as signed integer multiplication and the result will be cast to the size_t type. By defining the iterator as size_t the multiplication is handled as unsigned and then the result is stored as size_t.  The following screen shot shows the result of size_t * int vs (size_t) int * int:
Screen Shot 2022-01-13 at 11 47 44 AM

rouault: > If you do a cast after the multiplication

I meant before, like ``dirptr->filename_buf + (size_t)it_image * OPJ_PATH_LEN;``

Eharve14:
Forgive me for the repeated comment I haven't looked at this in a while and now that I am digging into it I am finding all of the issues that lead to the fix I tried to push originally:
Before the calloc there is an integer overflow in the get num images function:
`int get_num_images(char *imgdirpath)
{
    DIR *dir;
    struct dirent* content;
    int num_images = 0;

    /*Reading the input images from given input directory*/

    dir = opendir(imgdirpath);
    if (!dir) {
        fprintf(stderr, "Could not open Folder %s\n", imgdirpath);
        return 0;
    }

    while ((content = readdir(dir)) != NULL) {
        if (strcmp(".", content->d_name) == 0 || strcmp("..", content->d_name) == 0) {
            continue;
        }
        num_images++;
    }
    closedir(dir);
    return num_images;
}'

the maximum number of file in a dir in windows 10 is equal to SIZE_MAX, meaning that the signed integer operation will overflow if the number of files in the dir is >  2147483647. The integer will wrap and the calloc will allocate an amount of memory. 

The cast to size_t to fix the multiplication will still have an integer overflow error. 

The original solution prevents all of this by ensuring that num_images is < SIZE_MAX/(OPJ_PATH_LEN* 8) and I added a check to ensure num_images is not negative, resulting in unintended execution.

rouault:
> the maximum number of file in a dir in windows 10 is equal to SIZE_MAX, meaning that the signed integer operation will overflow if the number of files in the dir is > 2147483647

that's really a theoretical concern... In any case if we wanted to address this, this would need to be checked in get_num_images() before the overflow occurs

Eharve14:
I understand that the the concern is theoretical and unrealistic for normal use. It is however a very exploitable path and a real security vulnerability. 

It's the same exploitation path as the original CVE just with a larger number of files. Point to a dir with enough named files to create the overflow, use the file names to inject data onto the stack as the are read in past the end of the allocated memory.

Eharve14:
Should I move the check for num_images to the get_num_images function? Also are you suggesting the function should throw an error and exit? or are we adding a flag for overflow and checking the result?

rouault:
> Should I move the check for num_images to the get_num_images function? Also are you suggesting the function should throw an error and exit? or are we adding a flag for overflow and checking the result?

I'd prefer first that the scope of this PR is only to add the (size_t) casts so that we keep focus on one thing at a time

For another PR to address the overflow in get_num_images, I would just break from the for loop in get_num_images before the overflow of the counter occurs.

Eharve14:
Rodger that, I will fix this PR to just add the casts. I will open a separate pull request down the line for the get_num_images.

src/bin/jp2/opj_dump.c
```diff
@@ -529,13 +529,13 @@ int main(int argc, char *argv[])
         }
 
         for (it_image = 0; it_image < num_images; it_image++) {
-            dirptr->filename[it_image] = dirptr->filename_buf + it_image * OPJ_PATH_LEN;
+            dirptr->filename[it_image] = dirptr->filename_buf + (size_t)it_image * 
+                                          OPJ_PATH_LEN;
```
rouault, line 533:
```suggestion
                                         OPJ_PATH_LEN;

File: src/bin/jp2/opj_dump.c [hunk 3]
```diff
@@ -529,13 +529,13 @@ int main(int argc, char *argv[])
         }
 
         for (it_image = 0; it_image < num_images; it_image++) {
-            dirptr->filename[it_image] = dirptr->filename_buf + it_image * OPJ_PATH_LEN;
+            dirptr->filename[it_image] = dirptr->filename_buf + (size_t)it_image *
```
rouault, line 532:
```suggestion
            dirptr->filename[it_image] = dirptr->filename_buf + (size_t)it_image *
```

Eharve14:
Thanks for the assist, Sorry for spamming the commits.

PR #1397, opened 2022-01-13
Added check for integer overflow in get_num_images 

Eharve14:
As discussed in pull request 1396, added a check for integer overflow. 
Change list:
Defined num_images as unsigned int
Moved the if statement to check for an empty directory to the beginning of the read directory section
Added a check to see if num images rolls back to zero when incrementing.

### DIFF
--- a/src/bin/jp2/opj_compress.c
+++ b/src/bin/jp2/opj_compress.c
@@ -1957,6 +1957,11 @@ int main(int argc, char **argv)
     /* Read directory if necessary */
     if (img_fol.set_imgdir == 1) {
         num_images = get_num_images(img_fol.imgdirpath);
+        if((num_images > SIZE_MAX/(OPJ_PATH_LEN * sizeof(char))){
+            fprintf(stdout, "Max images exceeded\n");
+            ret = 0;
+            goto fin;
+        } else {
         dirptr = (dircnt_t*)malloc(sizeof(dircnt_t));
         if (dirptr) {
             dirptr->filename_buf = (char*)calloc(num_images, OPJ_PATH_LEN * sizeof(
@@ -1973,6 +1978,7 @@ int main(int argc, char **argv)
         if (load_images(dirptr, img_fol.imgdirpath) == 1) {
             ret = 0;
             goto fin;
+        }
         }
         if (num_images == 0) {
             fprintf(stdout, "Folder is empty\n");
--- a/src/bin/jp2/opj_compress.c
+++ b/src/bin/jp2/opj_compress.c
@@ -44,6 +44,7 @@
 #include <stdlib.h>
 #include <math.h>
 #include <assert.h>
+#include <limits.h>
 
 #ifdef _WIN32
 #include "windirent.h"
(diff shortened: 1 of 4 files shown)


src/bin/jp2/opj_compress.c
```diff
@@ -486,6 +486,9 @@ static unsigned int get_num_images(char *imgdirpath)
             continue;
         }
         num_images++;
+        if (num_images == 0) {
```
rouault, line 489: while this is technically correct, having behaviour that relies on overflow of unsigned types always makes me a bit nervous. I'd prefer a test on UINT_MAX before the incrementation is done (the advantage is that a build with  clang -fsanitize=undefined,unsigned-integer-overflow would also be happy). Same for below similar instances

src/bin/jp2/opj_decompress.c
```diff
@@ -374,7 +374,7 @@ int get_num_images(char *imgdirpath)
 {
     DIR *dir;
     struct dirent* content;
-    int num_images = 0;
+    unsigned int num_images = 0;
```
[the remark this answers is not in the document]
Eharve14: Okay, I will updated it.

src/bin/jp2/opj_compress.c
Eharve14: Revised as requested.

Eharve14:
@rouault I made the requested revisions. I was under the impression that unsigned overflow was always defined behavior. Thank you for the feedback.

src/bin/jp2/opj_compress.c
```diff
@@ -485,6 +486,9 @@ static unsigned int get_num_images(char *imgdirpath)
         if (strcmp(".", content->d_name) == 0 || strcmp("..", content->d_name) == 0) {
             continue;
         }
+        if (num_images == UINT_MAX) {
+            fprintf(stderr, "Too many files in folder %s\n", imgdirpath);
```
rouault, line 490: missing break;

src/bin/jp2/opj_decompress.c
```diff
@@ -388,7 +389,11 @@ int get_num_images(char *imgdirpath)
         if (strcmp(".", content->d_name) == 0 || strcmp("..", content->d_name) == 0) {
             continue;
         }
+        if (num_images == UINT_MAX) {
+            fprintf(stderr, "Too many files in folder %s\n", imgdirpath);
```
rouault, line 393: missing break;

src/bin/jp2/opj_dump.c
```diff
@@ -140,6 +141,9 @@ static int get_num_images(char *imgdirpath)
         if (strcmp(".", content->d_name) == 0 || strcmp("..", content->d_name) == 0) {
             continue;
         }
+        if (num_images == UINT_MAX) {
+            fprintf(stderr, "Too many files in folder %s\n", imgdirpath);
```
rouault, line 145: missing break;

src/bin/jp2/opj_compress.c
Eharve14: Added return statement

Eharve14:
Added the return statement after the overflow check. I don't know how I managed to mess up copy paste.

src/bin/jp2/opj_compress.c
rouault: if you use return, you need an explicit closedir(dir) before.
Other possibilities do:
num_images = 0;
break;
Same for other instances

Eharve14: Ah got you, for whatever reason I was thinking the goto fin statement in the calling section of code cleaned everything up, including the dir.  Also didn't realize the opendir function closed the directory if there was nothing there and freed the memory. I made the correction as requested. Thanks again.

kirotawa:
hey there, 

Is these fixes anyhow related with CVE-2021-29338 ? Should it have another CVE assigned to these fixes/commits in this merge?

Thanks!

PR #1398, opened 2022-01-16
Missed conversion from unsigned int to OPJ_INT32 

Eharve14:


Eharve14:
Accidentally deleted DWT.c thinking I was deleting the change. this pull request is to resolve a type conversion issue in the main function. The type for num_images in main is OPJ_INT32, return type of get_num_images is unsigned int.

Eharve14:
to be clear I reverted my branch to include DWT.c

src/bin/jp2/opj_decompress.c
```diff
@@ -475,7 +475,7 @@ const char* path_separator = "/";
 #endif
 
 /* -------------------------------------------------------------------------- */
-char get_next_file(int imageno, dircnt_t *dirptr, img_fol_t *img_fol,
+char get_next_file(unsigned int imageno, dircnt_t *dirptr, img_fol_t *img_fol,
```
rouault, line 478: fprintf(stderr, "File Number %d \"%s\"\n", imageno, image_filename); at line 486 should use "%u" instead of "%d". There might be other similar instances in the other files

PR #1402, opened 2022-01-18
More overflow related bug fixes

Eharve14:
Ran through the open security issues identified by LGTM code analysis and added fixes for the relevant issues.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "files.associations": {
+        "cstdlib": "c"
+    }
+}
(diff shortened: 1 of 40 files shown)


Eharve14 (commented):
The change from I to J was because the type of I is UINT_16, type of max is UINT_32. j is defined as UINT_32

rouault:
> The change from I to J was because the type of I is UINT_16, type of max is UINT_32. j is defined as UINT_32

ok, I don't think there was any issue in practice because the maximum number of components in a JPEG2000 file is 16384 and this is normally checked before

Eharve14:
If that check is always done, I agree that the change is not  necessary. I made the edit because that check is not in the function with the iterator in the loop in question, and I did not do the leg work to ensure the check is present before every call to this function. 

Also this is the only loop in that section of code with the type mismatch between the iterator and comparator.

PR #1403, opened 2022-01-23
opj_encoder_set_extra_options(): add a GUARD_BITS=value option

rouault:
and add a -GuardBits option to opj_compress.

The recently-released SMPTE DCP Bv2.1 Application Profile (link below)
says that the number of guard bits in the QCD marker shall be 1 for 2K
content and 2 for 4K content. This change allows the number of guard bits
to be configured, so that users of openjpeg have the control they need to meet the specification.

https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=9161348

This is an alternative implementation of https://github.com/uclouvain/openjpeg/pull/1388
that keeps ABI unchanged.


cth103:
Thank you! :+1:

PR #1416, opened 2022-03-25
Separate fuzz targets to increase coverage

Navidem:
As I checked locally, separating fuzz targets for J2K and JP2 leads to increased coverage.

I came across this case while testing our new tool [Fuzz-Introspector](https://github.com/ossf/fuzz-introspector) on OSS-Fuzz projects.

### DIFF
--- a/tests/fuzzers/opj_decompress_fuzzer_JP2.cpp
+++ b/tests/fuzzers/opj_decompress_fuzzer_JP2.cpp
@@ -112,7 +112,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *buf, size_t len)
 
     OPJ_CODEC_FORMAT eCodecFormat;
     if (len >= 4 + sizeof(jp2_box_jp) &&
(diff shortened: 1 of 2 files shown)


tests/fuzzers/opj_decompress_fuzzer_JP2.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+static const unsigned char jpc_header[] = {0xff, 0x4f};
+static const unsigned char jp2_box_jp[] = {0x6a, 0x50, 0x20, 0x20}; /* 'jP  ' */
+
+int LLVMFuzzerTestOneInput(const uint8_t *buf, size_t len)
+{
+
+    OPJ_CODEC_FORMAT eCodecFormat;
+    if (len >= 4 + sizeof(jp2_box_jp) &&
+               memcmp(buf + 4, jp2_box_jp, sizeof(jp2_box_jp)) == 0) {
```
rouault, line 115:
```suggestion
            memcmp(buf + 4, jp2_box_jp, sizeof(jp2_box_jp)) == 0) {
```

PR #1421, opened 2022-05-15
pkgconfig: Define OPJ_STATIC for static linking with pkgconf

Biswa96:
allows for the usage of $(pkgconf --static --cflags libopenjp2) to produce
the proper CFLAGS for static linking. Relies on pkgconf rather than pkg-config.

### DIFF
--- a/src/lib/openjp2/libopenjp2.pc.cmake.in
+++ b/src/lib/openjp2/libopenjp2.pc.cmake.in
@@ -12,3 +12,4 @@ Version: @OPENJPEG_VERSION@
 Libs: -L${libdir} -lopenjp2
 Libs.private: -lm
 Cflags: -I${includedir}
+Cflags.private: -DOPJ_STATIC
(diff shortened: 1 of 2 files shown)


Neumann-A:
This is also required for the CMake targets. Unfortunally openjpeg uses `add_definitions` instead of properly adding the definitions to the targets via `target_compile_definitions`. 

I currently see 
```
opj_jpip_test.c.obj : error LNK2019: unresolved external symbol __imp_get_index_from_JP2file referenced in function main
opj_jpip_test.c.obj : error LNK2019: unresolved external symbol __imp_destroy_index referenced in function main
opj_jpip_test.c.obj : error LNK2019: unresolved external symbol __imp_output_index referenced in function main
```
due to that

Biswa96:
Feel free to add that. I was only seeing the static build issue in ffmpeg which uses pkgconfig files. And this change fixes it.

PR #1424, opened 2022-05-24
Switch to GNUInstallDirs [v2]

laumann:
Rebase of #1165 on the current master, plus an attempt at moving docs `$DATAROOTDIR/doc/{OPENJPEG -> openjpeg}`. I considered alternatively to mangle CMAKE_INSTALL_DOCDIR directly, but I think that could cause other problems.


laumann:
cc @a17r @thesamesam

laumann:
@rouault ping, just to follow up on this, are you interested in this change?

rouault:
> just to follow up on this, are you interested in this change?

yes, but there are likely subtelies. In other projects where I'm involved with, it has been raised that sometimes CMAKE_INSTALL_{foo} is an absolute directory and not a relative directory (cf https://github.com/OSGeo/gdal/issues/6004 or https://github.com/OSGeo/PROJ/issues/3206), so I suspect that in your pull request CMAKE_INSTALL_FULL_{foo} should be used, at least for .pc file generation

laumann:
> > just to follow up on this, are you interested in this change?
> 
> yes, but there are likely subtelies. In other projects where I'm involved with, it has been raised that sometimes CMAKE_INSTALL_{foo} is an absolute directory and not a relative directory (cf [OSGeo/gdal#6004](https://github.com/OSGeo/gdal/issues/6004) or [OSGeo/PROJ#3206](https://github.com/OSGeo/PROJ/issues/3206)), so I suspect that in your pull request CMAKE_INSTALL_FULL_{foo} should be used, at least for .pc file generation

Sure, I'll look into it - it would make sense, in Gentoo's patching to use GNUInstallDirs the values for `mandir` and `docdir` omit the `${prefix}` as they already have the prefix. See: https://gitweb.gentoo.org/repo/gentoo.git/tree/media-libs/openjpeg/files/openjpeg-2.5.0-gnuinstalldirs.patch#n287

Would you prefer that `${prefix}` is removed from all the generated .pc entries? Ie, drop `${prefix}` and use the `CMAKE_INSTALL_FULL_{foo}` variants?

rouault:
> Ie, drop `${prefix}` and use the `CMAKE_INSTALL_FULL_{foo}` variants?

yes, my understanding is that it is the correct way

laumann:
> > Ie, drop `${prefix}` and use the `CMAKE_INSTALL_FULL_{foo}` variants?
> 
> yes, my understanding is that it is the correct way

OK, thanks for pointing it out. I'll make those changes (tomorrow probably, getting late here).

rouault:
> OK, thanks for pointing it out.

actually taking inspiration from https://github.com/OSGeo/PROJ/pull/3251/commits/ab25e4b7ed9544e668282dcd293cfaaa2e56dbdf which tests if CMAKE_INSTALL_{foo} is relative or not might better. The idea is to still use ${prefix}/${CMAKE_INSTALL_{foo}} when it is relative, so that packages can be relocated

laumann:
The reason I ask is that I'm considering one of two approaches:
 - Set variables like ${libdir}, ${includedir}, etc and replace all occurrences of ${CMAKE_INSTALL_LIBDIR} with ${libdir}, etc.
 - Focus on .pc file generation working correctly, where it takes into account whether or not CMAKE_INSTALL_LIBDIR (and others) are relative or full paths

I'm leaning towards the latter approach as it's less work. But I'd like to know which you prefer.

rouault:
> But I'd like to know which you prefer.

I'm not completely sure to have understood your first option. But IMHO ${libdir} shouldn't be used outside of .pc generation. ${CMAKE_INSTALL_LIBDIR} is idiomatic in all other places such as install() CMake function.
So your second point is more what I've in mind

laumann:
@rouault Sorry for taking a while to get back to this, I've implemented it now, let me know what you think. I had to use `\\\${prefix}` instead of `$\{prefix\}` not exactly sure why. I got an "invalid escape sequence \{" error with the latter format.

laumann:
@rouault Gentle ping on this, I think I've implemented what you requested.

rouault:
merged. thanks!

laumann:
Yay! Thank you!

arichardson:
This broke cross-compilation of downstream projects (poppler) for me, I've submitted #1439 to fix this.

SoapGentoo:
How does this break cross-compilation?

arichardson:
> How does this break cross-compilation?

I am installing using CMAKE_INSTALL_PREFIX=/usr/local and DESTDIR=/path/to/my/sysroot. Without the relocatable change, downstream projects will add `-I/usr/local` instead of `-I/path/to/my/sysroot/usr/local` and will not find the headers.

SoapGentoo:
That is correct, which is why cross-compilation generally expects `CMAKE_FIND_ROOT_PATH` or `CMAKE_SYSROOT`. [We use the former in Gentoo](https://gitweb.gentoo.org/repo/gentoo.git/tree/eclass/cmake.eclass#n491), and it works just fine. Said differently, if all cmake-based projects used garden path `GNUInstallDirs` as intended, none of them could be cross-compiled, which sounds counter-intuitive to what it tries to achieve?

arichardson:
> That is correct, which is why cross-compilation generally expects `CMAKE_FIND_ROOT_PATH` or `CMAKE_SYSROOT`. [We use the former in Gentoo](https://gitweb.gentoo.org/repo/gentoo.git/tree/eclass/cmake.eclass#n491), and it works just fine. Said differently, if all cmake-based projects used garden path `GNUInstallDirs` as intended, none of them could be cross-compiled, which sounds counter-intuitive to what it tries to achieve?

Yes, I set all of those, and that works fine for almost all projects (the ones that don't include those paths in the Config.cmake file, and instead use `configure_package_config_file` with `PATH_VARS`) or just link against the CMake targets instead of relying on FOO_INCS/FOO_LIBS (i.e. modern CMake).

SoapGentoo:
wouldn't it then make more sense to move to modern CMake where the user has to use `target_link_libraries` to pull in the correct target?

autoantwort:
This did not replace all usages of `OPENJPEG_INSTALL_INCLUDE_DIR` and breaks the generated cmake config files. 
https://github.com/uclouvain/openjpeg/blob/2d606701e8b7aa83f657d113c3367508e99bd12b/src/lib/openjp2/CMakeLists.txt#L108

thesamesam:
Please file a new bug then.

CMakeLists.txt
```diff
@@ -105,63 +105,32 @@ endif()
...
-  set(OPENJPEG_INSTALL_SHARE_DIR "share")
-endif()
-
-if(NOT OPENJPEG_INSTALL_DATA_DIR)
-  set(OPENJPEG_INSTALL_DATA_DIR "${OPENJPEG_INSTALL_SHARE_DIR}/${OPENJPEG_INSTALL_SUBDIR}")
-endif()
-
-if(NOT OPENJPEG_INSTALL_INCLUDE_DIR)
-  set(OPENJPEG_INSTALL_INCLUDE_DIR "include/${OPENJPEG_INSTALL_SUBDIR}")
-endif()
```
dg0yt, line 132: This change removes the ability to customize the include dir location in v2.5.1. This affects the vcpkg port. (I wouldn't care, but I learned that Visual Studio users feel less comfortable with versioned include paths if not using cmake.)

dg0yt: This would change the prefix (`include` -> `foo`) but not the suffix (`openjpeg-2.5`) because it hardcoded everywhere.

For the vcpkg port I decided to not add a patch (i.e. openjpeg decides), but write two forwarding headers to vcpkg's old location, marked as "legacy".
So probably no further change needed here - but a breaking change nevertheless.

rouault: @dg0yt If you want to contribute a pull request to fix that, that's a good moment to do because I'll likely have to issue a 2.5.2 because of https://github.com/uclouvain/openjpeg/issues/1514#issuecomment-1966373513

Issue #1413, opened 2022-03-03
opj_decompress heap overflow Denial of Service issue


smlijun:
## Expected behavior and actual behavior.

This vulnerability is a heap overflow bug issued by openjpeg opj_decompression.
This vulnerability can cause Denial of service.

## Steps to reproduce the problem.

The environment setting is as follows.
OS : ubuntu20.04
Build option : cmake .. -DCMAKE_BUILD_TYPE=Release (I follow build option in openjpeg Official github)
Download the latest version of openjpeg from git and build it as follows.
```bash
mkdir build
cd build
cmake .. -DCMAKE_BUILD_TYPE=Release
make
```
execute opj_decompress as follows
```bash
./opj_decompress -i payload.j2k -o test.pgm
```
Double free will occur

payload url : https://www.notion.so/openjpeg-opj_decompress-double-free-bug-report-ddb096dc92274d6b8e305efc3f931ac8#53973c09200c45ed9c03d0e2a91da87d

screen shot : https://caramel-abacus-14e.notion.site/openjpeg-opj_decompress-double-free-bug-report-ddb096dc92274d6b8e305efc3f931ac8

call stack
```
[#0](https://hackerone.com/reports/0) GI_raise (sig=sig[@entry](https://hackerone.com/entry)=0x6) at ../sysdeps/unix/sysv/linux/raise.c:50
[#1](https://hackerone.com/reports/1) 0x00007ffff7c0d859 in GI_abort () at abort.c:79
[#2](https://hackerone.com/reports/2) 0x00007ffff7c783ee in libc_message (action=action[@entry](https://hackerone.com/entry)=do_abort,
fmt=fmt[@entry](https://hackerone.com/entry)=0x7ffff7da2285 "%s\n") at ../sysdeps/posix/libc_fatal.c:155
[#3](https://hackerone.com/reports/3) 0x00007ffff7c8047c in malloc_printerr (
str=str[@entry](https://hackerone.com/entry)=0x7ffff7da4670 "double free or corruption (out)") at malloc.c:5347
[#4](https://hackerone.com/reports/4) 0x00007ffff7c82120 in _int_free (av=0x7ffff7dd3b80 <main_arena>, p=0x55555558abe0,
have_lock=<optimized out>) at malloc.c:4314
[#5](https://hackerone.com/reports/5) 0x00007ffff7fafbb4 in opj_aligned_free (ptr=0x55555558abf0)
[...]

PR #1426, opened 2022-05-31
Fix opj_t1_allocate_buffers malloc size error

zodf0055980:
Fix issue #1413

### DIFF
--- a/src/lib/openjp2/ht_dec.c
+++ b/src/lib/openjp2/ht_dec.c
@@ -1063,7 +1063,7 @@ static OPJ_BOOL opj_t1_allocate_buffers(
         if (flagssize > t1->flagssize) {
 
             opj_aligned_free(t1->flags);
-            t1->flags = (opj_flag_t*) opj_aligned_malloc(flagssize);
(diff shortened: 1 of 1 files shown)


src/lib/openjp2/ht_dec.c
```diff
@@ -1063,7 +1063,7 @@ static OPJ_BOOL opj_t1_allocate_buffers(
         if (flagssize > t1->flagssize) {
 
             opj_aligned_free(t1->flags);
-            t1->flags = (opj_flag_t*) opj_aligned_malloc(flagssize);
+            t1->flags = (opj_flag_t*) opj_aligned_malloc(flagssize * sizeof(opj_flag_t));
```
rouault, line 1066: at line 1074, the memset should also be changed to

memset(t1->flags, 0, flagssize * sizeof(opj_flag_t));

rouault (changes requested):

src/lib/openjp2/ht_dec.c
zodf0055980: You are right, I forget it.
I add this in new commit.

PR #867, opened 2016-11-14
Fix lcms2 static linking using pgk config

pseiderer:
thirdparty: lcms2: append flags found by pkg-config if available
    
This change allows to get all required CFLAGS/LDFLAGS in case of static only
build.
    
Fixes a buildroot build failure (see [1], [2] and [3]).
    
[1] http://autobuild.buildroot.net/results/5ce/5cee20afd8bef5268832cddcb3a5270746be7a57
[2] http://lists.busybox.net/pipermail/buildroot/2016-November/177187.html
[3] http://lists.busybox.net/pipermail/buildroot/2016-November/177188.html
    
Signed-off-by: Peter Seiderer <ps.report@gmx.net>
---
Depends on https://github.com/uclouvain/openjpeg/pull/866

PR #1431, opened 2022-06-27
Build: fix linking of executables on some systems where TIFF/LCMS2 static libraries are not in system directories (fixes #1430)

### LINKED ISSUE #1430: OSX m1 v2.5.0 build fail
## Expected behavior and actual behavior.

Expected: build completed.
Actual: build failed.

Error:

```
Linking C executable ../../../bin/opj_decompress
ld: library not found for -ltiff
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [bin/opj_decompress] Error 1
make[1]: *** [src/bin/jp2/CMakeFiles/opj_decompress.dir/all] Error 2
make: *** [all] Error 2
```

```bash
mkdir build
cd build
cmake .. -DCMAKE_BUILD_TYPE=Release
make
```

## Operating system

OSX 12.4 m1

## openjpeg version

v2.5.0

## Additional info

v2.4.0 - build works ok.

```bash
brew info libtiff
libtiff: stable 4.4.0 (bottled)
TIFF library and utilities
https://libtiff.gitlab.io/libtiff/
/opt/homebrew/Cellar/libtiff/4.4.0 (249 files, 4.8MB) *
```

rouault:
Note that the fix might be partial only for static-only builds (cf
comments)

Ammends PR #866 and #867

CC @ffontaine / @pseiderer

### DIFF
--- a/thirdparty/CMakeLists.txt
+++ b/thirdparty/CMakeLists.txt
@@ -89,8 +89,15 @@ else(BUILD_THIRDPARTY)
     message(STATUS "Your system seems to have a TIFF lib available, we will use it")
     set(OPJ_HAVE_TIFF_H 1 PARENT_SCOPE)
     set(OPJ_HAVE_LIBTIFF 1 PARENT_SCOPE)
-    set(TIFF_LIBNAME ${TIFF_LIBRARIES} ${PC_TIFF_STATIC_LIBRARIES} PARENT_SCOPE)
-    set(TIFF_INCLUDE_DIRNAME ${TIFF_INCLUDE_DIR} ${PC_TIFF_STATIC_INCLUDE_DIRS} PARENT_SCOPE)
+    if(BUILD_STATIC_LIBS AND NOT BUILD_SHARED_LIBS)
+      # Probably incorrect as PC_TIFF_STATIC_LIBRARIES will lack the path to the libraries
+      # and will only work if they are in system directories
+      set(TIFF_LIBNAME ${PC_TIFF_STATIC_LIBRARIES} PARENT_SCOPE)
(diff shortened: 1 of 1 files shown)


ffontaine:
I'm a little bit lost, `TIFF_LIBNAME` is set to  `${TIFF_LIBRARIES} ${PC_TIFF_STATIC_LIBRARIES}` which seems right except if `${PC_TIFF_STATIC_LIBRARIES}` is "broken"  on osx and contains a non existing path resulting in a build failure. Did I missed something?

rouault:
> except if `${PC_TIFF_STATIC_LIBRARIES}` is "broken" on osx and contains a non existing path resulting in a build failure.

PC_TIFF_STATIC_LIBRARIES on my Ubuntu 20.04 is "tiff;webp;zstd;lzma;jbig;jpeg;z;m" . Which is OK, but this variable doesn't contain paths, which causes the issue on distributions where the libraries are not in a system library directory. Presumably this should be combined with PC_TIFF_STATIC_LIBS_PATHS to build fully expanded library names.

rouault:
> I fear that it will break the build of `openjp2_static` when `BUILD_SHARED_LIBS` and `BUILD_STATIC_LIBS` are both set to `ON`

yes that's likely. Perhaps do 2 separate builds. Static builds with dependency tracking are quite difficult to get right, and openjpeg that attempts to do both at the same time make it even more challenging.

ffontaine:
Indeed, we should definitely remove the `BR2_SHARED_STATIC_LIBS` option in buildroot.
Thanks for pinging me about this change. I'll be able to correctly handle the bump of openjpeg in buildroot.

PR #1439, opened 2022-08-11
Make OpenJPEGConfig.cmake relocatable with CMake > 3.0

arichardson:
Using CMakePackageConfigHelpers, we can generate a relocatable
OpenJPEGConfig.config, using the PATH_VARS feature to make
CMAKE_INSTALL_LIBDIR relative to the installed location.
This change is needed for me when cross-compiling since
CMAKE_INSTALL_FULL_LIBDIR is a path inside the sysroot rather than
an absolute path to the actual includes. Without this change poppler
ends up passing a -I flag that does not exist.

This includes fallback code for CMake 2.8, which adds a bit of complexity,
since I'm not sure if raising the minimum to 3.0 (now over 8 years old)
is acceptable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -304,12 +304,23 @@ if(BUILD_TESTING)
 endif()
 
 #-----------------------------------------------------------------------------
-# install all targets referenced as OPENJPEGTargets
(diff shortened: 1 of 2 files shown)


rouault:
thanks

PR #1441, opened 2022-08-12
Fix Heap-buffer-overflow READ in opj_jp2_apply_pclr

sashashura:
The issue was found while fuzzing opencv:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=47342

The read overflow triggered by reading `src[j]` in
```cpp
            for (j = 0; j < max; ++j) {
                dst[j] = src[j];
            }
```
The max is calculated as `new_comps[pcol].w * new_comps[pcol].h`, however the `src = old_comps[cmp].data;` which may have different `w` and `h` dimensions.

### DIFF
--- a/src/lib/openjp2/jp2.c
+++ b/src/lib/openjp2/jp2.c
@@ -1042,7 +1042,7 @@ static OPJ_BOOL opj_jp2_apply_pclr(opj_image_t *image,
     OPJ_UINT32 *entries;
     opj_jp2_cmap_comp_t *cmap;
     OPJ_INT32 *src, *dst;
-    OPJ_UINT32 j, max;
+    OPJ_UINT32 j, max, newmax, oldmax;
     OPJ_UINT16 i, nr_channels, cmp, pcol;
     OPJ_INT32 k, top_k;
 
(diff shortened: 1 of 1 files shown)


src/lib/openjp2/jp2.c
```diff
@@ -1108,7 +1108,10 @@ static OPJ_BOOL opj_jp2_apply_pclr(opj_image_t *image,
         pcol = cmap[i].pcol;
         src = old_comps[cmp].data;
         assert(src); /* verified above */
-        max = new_comps[pcol].w * new_comps[pcol].h;
+        oldmax = old_comps[cmp].w * old_comps[cmp].h;
```
rouault, line 1111: shouldn't the fix to be just:
``max = new_comps[i].w * new_comps[i].h;``

sashashura: It can be... While I understand where and why it accesses out of bands memory, my assumption about the fix can be wrong because I'm not so familiar with the code base. I have applied and tested your suggestion and the crash is gone.

sashashura:
Hi, do you have an estimate when 2.5.1 will be released? Thanks!

rouault:
> Hi, do you have an estimate when 2.5.1 will be released? Thanks!

likely not before several months

PR #1456, opened 2023-02-06
Fix #1424

autoantwort:


autoantwort:
I don't see why a test should fail because of my change?

rouault:
> I don't see why a test should fail because of my change?

they are likely unrelated. Your change looks good to me

PR #1458, opened 2023-02-24
Add loongarch 256-bit LASX SIMD optimization

jinboson:
With the bench_dwt utility, on loongarch64:

before changes: 30.990 s
with LASX optimization: 3.334 s

### DIFF
--- a/src/lib/openjp2/t1.c
+++ b/src/lib/openjp2/t1.c
@@ -1786,6 +1786,34 @@ static void opj_t1_clbl_decode_processor(void* user_data, opj_tls_t* tls)
                 }
             }
 #endif
+
+#ifdef __loongarch_asx
+            {
+                asm volatile
+                (
+                    "srli.w        $t0,     %[cblk],   4      \n\t"
+                    "xvldrepl.w    $xr0,    %[step],   0      \n\t"
+                    "beqz          $t0,     999f              \n\t"
+                    "16:                                      \n\t"
+                    "xvld          $xr1,    %[data],   0      \n\t"
+                    "xvld          $xr2,    %[data],   32     \n\t"
+                    "xvffint.s.w   $xr1,    $xr1              \n\t"
+                    "xvffint.s.w   $xr2,    $xr2              \n\t"
+                    "xvfmul.s      $xr1,    $xr1,      $xr0   \n\t"
+                    "xvfmul.s      $xr2,    $xr2,      $xr0   \n\t"
+                    "xvst          $xr1,    %[data],   0      \n\t"
+                    "xvst          $xr2,    %[data],   32     \n\t"
+                    "addi.d        %[data], %[data],   64     \n\t"
+                    "addi.w        $t0,     $t0,       -1     \n\t"
+                    "bnez          $t0,     16b               \n\t"
+                    "999:                                     \n\t"
+                    : [data]"+r"(datap)
+                    : [step]"r"(&stepsize), [cblk]"r"(cblk_size)
+                    : "$t0", "$xr0", "$xr1", "$xr2", "memory"
+                );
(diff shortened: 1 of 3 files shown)


src/lib/openjp2/t1.c
```diff
@@ -1786,6 +1786,34 @@ static void opj_t1_clbl_decode_processor(void* user_data, opj_tls_t* tls)
                 }
             }
 #endif
+
+#ifdef __loongarch_asx
+            {
+                asm volatile
```
xry111, line 1792: Using "volatile" is generally a bad idea (unless there is a side-effect not explainable by the input/output/clobber).  If it does not work w/o "volatile", you've likely missed some register input/output/clobber.

src/lib/openjp2/t1.c
```diff
@@ -1786,6 +1786,34 @@ static void opj_t1_clbl_decode_processor(void* user_data, opj_tls_t* tls)
...
+                (
+                    "srli.w        $t0,     %[cblk],   4      \n\t"
+                    "xvldrepl.w    $xr0,    %[step],   0      \n\t"
```
xry111, line 1795: Is there any reason not to use `xvreplgr2vr.w` here?

src/lib/openjp2/t1.c
```diff
@@ -1797,6 +1825,34 @@ static void opj_t1_clbl_decode_processor(void* user_data, opj_tls_t* tls)
                                                        (OPJ_SIZE_T)x];
         for (j = 0; j < cblk_h; ++j) {
             i = 0;
+#ifdef __loongarch_asx
+            {
+                OPJ_INT32* ptr1 = datap  + j * cblk_w;
+                OPJ_INT32* ptr2 = tiledp + j * (OPJ_SIZE_T)tile_w;
+                OPJ_UINT32 step_size = 0;
+                asm volatile
```
xry111, line 1833: Likewise, try to avoid `volatile`.

src/lib/openjp2/t1.c
jinboson: affect performance or something else ?

src/lib/openjp2/t1.c
jinboson: Nice catch, will update.

src/lib/openjp2/t1.c
xry111: Because "if it's a chicken, don't model it as a duck".

jinboson: Please , provide helpful suggestions.

xry111: Is it broken after if we remove `volatile`?  If it's not broken, just remove `volatile`.  Otherwise try to figure out why the compiler breaks it and fix the input/output/clobber list.

`volatile` **means** there is some unexplainable side effect, for example writing/reading some CSRs, invoking hardware memory barriers, etc.  These just do not apply for a SIMD optimization.

jinboson: just want to tell compiler not to optimize or modify the asm block.

xry111: > just want to tell compiler not to optimize or modify the asm block.

It won't modify the asm block because the compiler does not parse asm at all.

The only possible optimizations are:

1. Reorder the asm block with the instructions generated by the compiler itself.
2. Remove the asm block completely.

If the input/output/clobber lists are correct, 2 should be disabled, and 1 should be performed in a constrained way not to break the code.

xry111: Well, and maybe

3. Duplicating the asm block due to a loop unrolling etc.  Again if the input/output/clobber lists are correct this should happen in a safe way.

PR #1462, opened 2023-03-09
CMake: error out on warnings for strict/missing prototypes.

sebras:
And fix strict-prototypes/missing-prototypes warnings.


rouault:
@sebras errors for the mingw_64 target are at https://my.cdash.org/viewBuildError.php?buildid=2301396

sebras:
> errors for the mingw_64 target are at https://my.cdash.org/viewBuildError.php?buildid=2301396

I know, I found them. I've fixed one of them, but the error relating to DllMain is puzzling to me (I'm mainly a Linux developer). :)

rouault:
perhaps try the following change:
```
diff --git a/src/lib/openjp2/openjpeg.c b/src/lib/openjp2/openjpeg.c
index 29d3ee52..561e603d 100644
--- a/src/lib/openjp2/openjpeg.c
+++ b/src/lib/openjp2/openjpeg.c
@@ -33,6 +33,7 @@
 
 #ifdef _WIN32
 #include <windows.h>
+#include <process.h>
 #endif /* _WIN32 */
 
 #include "opj_includes.h"
```

sebras:
I'll just follow the lead of cairo and declare DllMain locally in the file.
This _should_ avoid the warning/error, and I can't forsee that it would create any issues.

PR #1470, opened 2023-06-29
Set a threshold for allocated memory in the opj_malloc file

pic4xiu:
When I was fuzzing, I found a file that can bypass all current error checks. This file can cause program denial of service, similar to cve-2019-6988. 

The poc is [here](https://github.com/pic4xiu/pocRep/blob/main/poc)

Run: `opj_decompress -i poc -o te.png`

I used gdb to debug and trace, and found that the program will call opj_calloc to open 2g of memory in each large loop, and this poc will cause the program to loop thousands of times. 

Perhaps it is more appropriate to introduce a new parameter to allow users to choose how many resources can be provided, but this will be an API change, and the impact will be great. So I directly introduced a variable for statistical memory in the function. In fact, similar functions can also be counted into variables, but because I have not observed a situation that can trigger similar bugs, so this commit does not add changes to other functions.

### DIFF
--- a/src/lib/openjp2/opj_malloc.c
+++ b/src/lib/openjp2/opj_malloc.c
@@ -197,6 +197,16 @@ void * opj_malloc(size_t size)
 }
 void * opj_calloc(size_t num, size_t size)
 {
+    static unsigned long long allocated_size = 0;
+    static unsigned long long max_allocated_size = 4ULL * 1024 * 1024 * 1024;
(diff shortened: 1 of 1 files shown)


pic4xiu:
Does the threshold need to be set larger? I didn't expect the test case to be so big

bastien-roucaries:
Why closing the fix for CVE-2023-39328 ??

PR #1479, opened 2023-09-07
ht_dec.c: Improve MSVC arm64 popcount performance

PeterJohnson:
Use NEON instructions for ARM64 (implementation based on microsoft/STL#2127).

Godbolt output here: https://godbolt.org/z/q7GPTqT14


asmorkalov:
Friendly reminder.

rouault:
Is testing for defined(OPJ_COMPILER_MSVC) && defined(_M_ARM64) sufficient to guarantee Neon instructions to be available ?
https://github.com/microsoft/STL/pull/2127/files has much more conditions

PR #1493, opened 2023-12-06
Add cmake version file

domin144:
Add the missing cmake version file, so that software using the library can check for version compatibility.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -296,9 +296,15 @@ configure_package_config_file(${CMAKE_CURRENT_LIST_DIR}/cmake/OpenJPEGConfig.cma
   ${OPENJPEG_BINARY_DIR}/OpenJPEGConfig.cmake
   INSTALL_DESTINATION ${OPENJPEG_INSTALL_PACKAGE_DIR}
   PATH_VARS CMAKE_INSTALL_INCLUDEDIR)
-install( FILES ${OPENJPEG_BINARY_DIR}/OpenJPEGConfig.cmake
-  DESTINATION ${OPENJPEG_INSTALL_PACKAGE_DIR}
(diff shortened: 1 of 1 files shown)


domin144:
cmake/OpenJPEGConfigVersion.cmake.in is copied from cmake sources, covered by 3-clause BSD license.
If this creates a licensing issue, the change could be restricted to cmake > 3.0.0 only.

They have no copyright header and cmake documentation says "they can be used as starting point to create more sophisticated custom ConfigVersion.cmake files"[1], so not sure if we should bother.

[1] https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html

rouault:
@mayeut does that look good to you ?

domin144:
One more doubt. I used `SameMajorVersion` option. If hypothetical version 3 is supposed to be compatible with any previous version, than `AnyNewerVersion` should be used instead (and BasicConfigVersion-AnyNewerVersion.cmake.in in place of BasicConfigVersion-SameMajorVersion.cmake.in).

rouault:
> If the current version is API compatible with 1.x,

it is not

rouault:
I believe we should just remove support for CMake < 3.0. Newer CMake versions are actually going to remove support for CMake < 3.5 soon. That way we could remove cmake/OpenJPEGConfigVersion.cmake.in

domin144:
If dropping support for old cmake is an option, I would be really happy with that.
I think this should be a separate PR and this PR could be rebased then.

rouault:
> If dropping support for old cmake is an option, I would be really happy with that.

would you be interested in tackling that?

> I think this should be a separate PR and this PR could be rebased then.

indeed

PR #1519, opened 2024-02-29
Fix line endings (CRLF -> LF)

stweil:


stweil:
I did not update a file from thirdparty/liblcms2 (which might be updated from 2.8 to 2.16).

PR #1520, opened 2024-03-02
Fix building on OpenBSD big endian hosts

brad0:
There is a swap16() function which conflicts with such a macro
in the endian.h header.  uint16_t swap16(uint16_t val16);

### DIFF
--- a/src/bin/jp2/convert.c
+++ b/src/bin/jp2/convert.c
@@ -664,7 +664,7 @@ static int tga_readheader(FILE *fp, unsigned int *bits_per_pixel,
 
 #ifdef OPJ_BIG_ENDIAN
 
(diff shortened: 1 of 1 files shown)


rouault:
> There is a swap16() function which conflicts with such a function
> in the endian.h header

if it is a function, then !defined(swap16) will not detect it, unless it is actually a macro?

a more robust fix would be to rename the function in openjpeg as opj_swap16

brad0:
> > There is a swap16() function which conflicts with such a function
> > in the endian.h header
> 
> if it is a function, then !defined(swap16) will not detect it, unless it is actually a macro?
> 
> a more robust fix would be to rename the function in openjpeg as opj_swap16

Ya, I was just going to post the other option was to rename the function in openjpeg, which is what most projects tend to do. Yes, in the header it is a macro.

PR #1529, opened 2024-05-15
Support setting enumcs for CMYK and EYCC color space

radarhere:
As per
https://github.com/uclouvain/openjpeg/blob/70e6263705334f854a27340e34ede11a767918ed/src/lib/openjp2/jp2.c#L2874-L2884
adds CMYK and EYCC to
https://github.com/uclouvain/openjpeg/blob/70e6263705334f854a27340e34ede11a767918ed/src/lib/openjp2/jp2.c#L1992-L1998

### DIFF
--- a/src/lib/openjp2/jp2.c
+++ b/src/lib/openjp2/jp2.c
@@ -1995,6 +1995,10 @@ OPJ_BOOL opj_jp2_setup_encoder(opj_jp2_t *jp2,
             jp2->enumcs = 17;    /* greyscale */
         } else if (image->color_space == 3) {
             jp2->enumcs = 18;    /* YUV */
+        } else if (image->color_space == 4) {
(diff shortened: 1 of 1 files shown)


src/lib/openjp2/jp2.c
```diff
@@ -1995,6 +1995,10 @@ OPJ_BOOL opj_jp2_setup_encoder(opj_jp2_t *jp2,
             jp2->enumcs = 17;    /* greyscale */
         } else if (image->color_space == 3) {
             jp2->enumcs = 18;    /* YUV */
+        } else if (image->color_space == 4) {
```
rouault, line 1998: would you mind using the OPJ_CLRSPC_xxxx enumeration values (for those new entries as well as for the 3 above existing ones?)

PR #1555, opened 2024-09-19
bench_dwt: Add assert for memory allocation failure

hleft:
Fix memory allocation failure on K230


hleft:
The K230 is a RISC-V device with only 500MB of memory, but the allocation can succeed on the Banana F3 (RISC-V 4GB)

PR #1562, opened 2024-11-17
Build: compile source files in parallel under MSVC

aras-p:
When using cmake with Visual Studio project files (default on Windows), add MSVC compile flag that enables using more than 1 CPU core during the build (i.e. compile source files in parallel).

On my PC (Ryzen 5950X), full rebuild of the MSVC solution in Release config goes from 4.6sec down to 2.4sec. Not much, but there's no downside of using the `/MP` flag. (I've recently landed the same build improvement in [OpenEXR](https://github.com/AcademySoftwareFoundation/openexr/pull/1858) and [OpenColorIO](https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/2072))


rouault:
I'm perplexed by this. Doesn't standard CMake options provide that capability already? If not, then this should be something to enhance on CMake itself, instead of hacking every software here and there

aras-p:
> I'm perplexed by this. Doesn't standard CMake options provide that capability already? If not, then this should be something to enhance on CMake itself, instead of hacking every software here and there

There's a 7 year old open issue in cmake project about this yes: https://gitlab.kitware.com/cmake/cmake/-/issues/16951
And no, cmake of today still does not do it out of the box, so "add this manually to every project out there" it is :(

rouault:
Closing. OpenJPEG takes just a few seconds to build. Not worth complicating our build script

PR #1572, opened 2024-12-26
If less than 2 bytes left, there will not be another SOT or an EOC

radarhere:


rouault:
@radarhere Can you attach a JPEG2000 file demonstrating the need for this change ?

radarhere:
Oh, I didn't mean to imply there was a need. I just considered `opj_j2k_read_sod` in isolation, and thought that if there were less than 2 bytes left, then
https://github.com/uclouvain/openjpeg/blob/eb25a5ec777ff6699f4bb1187740467dcfa64dd6/src/lib/openjp2/j2k.h#L130
wasn't the case, and
https://github.com/uclouvain/openjpeg/blob/eb25a5ec777ff6699f4bb1187740467dcfa64dd6/src/lib/openjp2/j2k.h#L133
reflected reality better.

If you wish to decline this for the sake of performance or for the sake of code simplicity, that is fine.

rouault:
IMHO this code base is sufficiently tricky to avoid adding new code without a clear evidence it is needed

radarhere:
Ok, sure. Thanks for the review.

PR #1573, opened 2025-01-16
opj_jp2_read_header: Check for error after parsing header.

sebras:
Consider the case where the caller has not set the p_image pointer to NULL before calling opj_read_header().

If opj_j2k_read_header_procedure() fails while obtaining the rest of the marker segment when calling opj_stream_read_data() because the data stream is too short, then opj_j2k_read_header() will never have the chance to initialize p_image, leaving it uninitialized.

opj_jp2_read_header() will check the p_image value whether opj_j2k_read_header() suceeded or failed. This may be detected as an error in valgrind or ASAN.

The fix is to check whether opj_j2k_read_header() suceeded before using the output argument p_image.

### DIFF
--- a/src/lib/openjp2/jp2.c
+++ b/src/lib/openjp2/jp2.c
@@ -2873,7 +2873,7 @@ OPJ_BOOL opj_jp2_read_header(opj_stream_private_t *p_stream,
                               p_image,
                               p_manager);
 
-    if (p_image && *p_image) {
+    if (ret && p_image && *p_image) {
         /* Set Image Color Space */
         if (jp2->enumcs == 16) {
             (*p_image)->color_space = OPJ_CLRSPC_SRGB;


sebras:
In PR #1463 in commit 0f528e95788863608aa1772f5370659edf618793, @rouault accidentally forgot to check the return value of opj_j2k_read_header() before processing its output arguments in opj_jp2_read_header(). This onliner PR fixes that minor buglet.

Thanks for maintaining openjpeg! :)

rouault:
thanks

opoplawski:
This apparently has been assigned CVE-2025-54874.  I'm trying to determine if this applies to openjpeg 2.0.  In that version there is no extra processing after the call to opj_j2k_read_header, so I'm inclined to believe that this was introduced in 2.5.1 with the mentioned commit.  The CVE report just says "2.5.3 and earlier".

rouault:
> I too would like to know the actual affected versions

yes 2.5.1, 2.5.2, 2.5.3

calixteman:
@rouault would it be possible to have a new release containing with this fix ?
I can cherry pick the patch but for sure a release would be better.
Thank you.

PR #1574, opened 2025-01-24
Fixed ICC profile copy failure on write

mircomir:
When I set a color profile in write I always get an assert in `opj_jp2_write_colr()` (line 831). I've been using the library for a short time, but I didn't find any other way to pass the profile on write. Looking at the code it seems that the color profile set in the image is never copied to `jp2->color`.


rouault:
@mircomir can you fix formatting issues ? See https://github.com/uclouvain/openjpeg/actions/runs/12952828902/job/36315685711?pr=1574

PR #1584, opened 2025-04-25
Add support for Apple framework builds

Treata11:


Treata11:
Greetings,

In I introduced the required properties for framework generation as described in [CMake documentation](https://cmake.org/cmake/help/latest/prop_tgt/FRAMEWORK.html#prop_tgt:FRAMEWORK)

I built the project with the following:

```cmake
cmake -S. -B build_iOS \
-DBUILD_SHARED_LIBS=OFF \
-DBUILD_CODEC=OFF \
-DOPJ_FRAMEWORK=ON \
-DCMAKE_SYSTEM_NAME=iOS

sudo cmake --build build_iOS --target install --config Release
```

And the framework artifact is created at `build_iOS/bin/openjp2.framework` with `openjpeg.h` & `opj_config.h` as the public headers & also some resources.

rouault:
I've tried  this PR on Linux and "make install" fails with:
```
CMake Error at cmake_install.cmake:80 (file):
  file INSTALL cannot find
  "/home/even/openjpeg/openjpeg/build/OpenJPEGTargets.cmake": No such file or
  directory.
```

PR #1596, opened 2025-09-26
Fix pkgconfig 'deps' generation

### LINKED ISSUE #1595: Incorrect .pc file generate for clang64 with msys2
clang64 in msys2 uses clang+llvm+ucrt for targetting Windows.
The generated .pc file looks like:
```sh
$ cat /clang64/lib/pkgconfig/libopenjp2.pc
prefix=/clang64
bindir=${prefix}/bin
mandir=${prefix}/share/man
docdir=${prefix}/share/doc/openjpeg
libdir=${prefix}/lib
includedir=${prefix}/include/openjpeg-2.5

Name: openjp2
Description: JPEG2000 library (Part 1 and 2)
URL: http://www.openjpeg.org/
Version: 2.5.4
Libs: -L${libdir} -lopenjp2
Libs.private: -l-lpthread
Cflags: -I${includedir}
Cflags.private: -DOPJ_STATIC
```

Note the Libs.private.

For reference, the gcc based environments have an empty Libs.private.
This could be due to gcc implicitly linking pthreads with mingw-w64, and thus cmake didn't include it implitcly.

Looks to be due to https://github.com/uclouvain/openjpeg/commit/d42966d0078245b2d598407739fd538d3d1f6833

Downstream issue https://github.com/msys2/MINGW-packages/issues/25687 and build file https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-openjpeg2/PKGBUILD

dg0yt:
Export `-pthread` as is. Fixes #1595.
Cleanup.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -332,10 +332,14 @@ set_variable_from_rel_or_absolute_path("includedir" "\\\${prefix}" "${CMAKE_INST
 function(get_pkgconfig_deps NAME OUT)
   get_target_property(link_libs ${NAME} LINK_LIBRARIES)
   if(NOT link_libs STREQUAL "link_libs-NOTFOUND")
-    foreach(lib ${link_libs})
-      # omit -lopenjp2 from Libs.private because libopenjp2 is in Requires
-      if(NOT ${lib} STREQUAL ${OPENJPEG_LIBRARY_NAME})
-        string(APPEND deps "-l${lib} ")
+    # omit openjp2 from Libs.private because libopenjp2 is in Requires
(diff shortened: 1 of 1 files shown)


1480c1:
As far as I can tell, this would not actually fix the linked issue as the flag is `-lpthread` in that case, not `-pthread`.

I've seen some other cmake projects that instead does two checks, first for if it's an absolute path, then one for if it looks like a flag (either starts with `-` or they check a bit more specific on `-l` vs `-f` vs no prefix). I'm not sure

dg0yt:
Wait, I see what you mean.

PR #1603, opened 2025-11-21
Fix parser hang when EOC marker appears before any SOT marker

mariusherzog:
Under certain circumstances, an invalid JPEG2000 file, ie. when an EOC marker is encountered before any SOT marker, leads to an endless loop when trying to parse the header in opj_read_header.
Attached is a file where this error occurs.
[invalid.debug](https://github.com/user-attachments/files/23673702/invalid.debug)


rouault:
> Attached is a file where this error occurs.

I can't reproduce any hang with latest openjpeg master:
```
$ ./bin/opj_decompress  -i ~/Téléchargements/invalid.debug -o out.ppm

===========================================
The extension of this file is incorrect.
FOUND ebug. SHOULD BE .j2k or .jpc or .j2c or .jhc
===========================================

[INFO] Start to read j2k main header (0).
[WARNING] Unknown marker
[INFO] Stream reached its end !
[ERROR] Stream too short
[ERROR] Unknown marker has been detected and generated error.
ERROR -> opj_decompress: failed to read the header
```

mariusherzog:
There was a subtle bug in my opj_stream_set_read_function which did not correctly detect the end-of-stream in this case. Sorry for the inconvenience.

PR #1629, opened 2026-04-12
NEON-optimize 9-7 IDWT

nico:
Takes `bin/bench_dwt -I` from 0.865 s to 0.672 s on my system.

---

"My system" being a Apple M4 Max MacBook Pro.

This code is hot when rendering https://archive.org/details/disquisitionesa00gaus in pdfium. This speeds up `pdfium_test --pages=1-50 ~/Downloads/disquisitionesa00gaus.pdf` by ca 10%.

This is similar to the existing SSE fast path.

PR #1630, opened 2026-04-12
NEON-optimize 5-3 IDWT

nico:
Takes `bin/bench_dwt` from 1.618 s to 0.432 s on my system.

---

Sadly, despite being a much bigger win in `bench_dwt` time, it's a much smaller end-to-end time improvement than #1629, as time spent in idwt is smaller than time spent in T1 / MQC decoding for lossless files (as there's more data). But still, it seems nice to have fast 5-3 IDWT anyways.

For a large lossless file I have, it takes `bin/opj_decompress -i balloon-reversible.jp2 -o test.ppm -threads 0` from printing "decode time: 588 ms" to "decode time: 559 ms", around a 5% speedup for decoding. It's not nothing, but much less than the other PR.

I created that input file with `bin/opj_compress -i image.ppm -o balloon-reversible.jp2 -M 1`, where image.ppm is balloon.jp2 in decompressed (and balloon.jp2 is the usual jp2 balloon test file).

### DIFF
--- a/src/lib/openjp2/dwt.c
+++ b/src/lib/openjp2/dwt.c
@@ -55,6 +55,9 @@
 #if (defined(__AVX2__) || defined(__AVX512F__))
 #include <immintrin.h>
 #endif
+#ifdef __ARM_NEON
+#include <arm_neon.h>
+#endif
 
 #if defined(__GNUC__)
 #pragma GCC poison malloc calloc realloc free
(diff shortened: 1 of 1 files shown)


nico:
For this one, I also verified that `ctest --parallel` has the same 50 failures it was without the PR, and I checked that dwt.c compiles fine with `--target=armv7a-linux-gnueabihf --sysroot ~/src/chrome/src/build/linux/debian_bullseye_armhf-sysroot` (to test 32-bit arm).

PR #1632, opened 2026-04-24
Fix a few warnings

nico:
With this, things build without warnings for me (macOX, Xcode 26.2) when using
CMAKE_C_FLAGS="-Wall -Wextra -Wconversion -Wno-unused-parameter -Wdeclaration-after-statement -Werror=declaration-after-statement"

Per https://github.com/uclouvain/openjpeg/pull/1629#issuecomment-4309230241, those are the warnings used on CI.

No intended behavior change.

### DIFF
--- a/src/lib/openjp2/mqc.c
+++ b/src/lib/openjp2/mqc.c
@@ -370,11 +370,6 @@ void opj_mqc_erterm_enc(opj_mqc_t *mqc)
     }
 }
 
-static INLINE void opj_mqc_renorme(opj_mqc_t *mqc)
-{
-    opj_mqc_renorme_macro(mqc, mqc->a, mqc->c, mqc->ct);
-}
-
 /**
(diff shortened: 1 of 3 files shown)


nico:
This should be harmless to merge, as it's behavior preserving.

nico:
Hey, the first commit effectively cherry-picks https://source.chromium.org/chromium/chromium/src/+/main:third_party/pdfium/third_party/libopenjpeg/0039-opj_mqc_renorme.patch;l=1?q=0039-opj_mqc_renorme.patch&sq= , so this is in a way part of upstreaming pdfium downstream patches. (I wasn't aware of that when I made this PR.)

nico:
@rouault Friendliest, gentlest, tendermost ping.

rouault (approved):

nico:
Thanks!

Issue #1631, opened 2026-04-21
Bug Report: invalid decoded-resolution state causes READ SEGV in opj_tcd_get_decoded_tile_size()


5asever40-a11y:
## Summary

A malformed JPEG 2000 codestream can reproducibly crash the OpenJPEG tile header decoding path under AddressSanitizer.

The crash occurs as a READ SEGV in:

```text
opj_tcd_get_decoded_tile_size()
src/lib/openjp2/tcd.c:1426
```

I am not making a strong claim about code execution. At minimum, this appears to be a reproducible decoder crash / robustness issue caused by invalid decoded-resolution state handling.

## Reproducer

Attached archive:

[openjpeg_tcd1426.zip](https://github.com/user-attachments/files/26921644/openjpeg_tcd1426.zip)

The archive contains:

```text
OpenJPEG_tcd1426_invalid_resolution_state_report_en.md
openjpeg_tcd1426_crash_input.j2k
openjpeg_tcd1426_asan.log
```

Crash input metadata:

```text
Filename: openjpeg_tcd1426_crash_input.j2k
SHA-256: 660ce4995e12714ba9ed3b19b310fd3d52a5bff27e42a3bf3690bba976083f2e
Size: 5476 bytes
```

## Reproduction Command

```bash
BIN=/home/js/ctf_dreamhack/fuzzing/2_fuzz/openjpeg/campaign/bin/tile_header_data_partial_afl_asan
CRASH="/path/to/openjpeg_tcd1426_crash_input.j2k"

ASAN_OPTIONS=detect_leaks=0:symbolize=1:handle_segv=1:abort_on_error=0 \
UBSAN_OPTIONS=print_stacktrace=1:halt_on_error=1 \
"$BIN" "$CRASH" 2>&1 | tee /tmp/openjpeg_tcd1426_asan.log
```

## ASAN Output

```text
AddressSanitizer:DEADLYSIGNAL
=================================================================
ERROR: AddressSanitizer: SEGV on unknown address 0x615fffffffc8
The signal is caused by a READ memory access.
    #0 opj_tcd_get_decoded_tile_size
       src/lib/openjp2/tcd.c:1426:37
    #1 opj_j2k_read_tile_header
       src/lib/openjp2/j2k.c:10097:24
    #2 opj_read_tile_header
       src/lib/openjp2/openjpeg.c:575:16
    #3 LLVMFuzzerTestOneInput
[...]

PR #1640, opened 2026-05-30
opj_tcd: fix out-of-bounds read in opj_tcd_get_decoded_tile_size() when the reduce factor equals numresolutions (#1631)

Alearner12:
## PR: Fix OOB read in opj_tcd_get_decoded_tile_size() when reduce factor == numresolutions

**Base:** `21b70b0` (v2.5.4-23) · **Branch:** `fix-1631-resolution-factor-oob` · **Diff:** j2k.c +12/-2, tcd.c +7/-1
**Closes:** #1631 (reported by @5asever40-a11y, crediting OH HAN GUEL and KANG DAEUN)

---

### Bug
`opj_tcd_get_decoded_tile_size()` (tcd.c:1421) indexes
`l_tile_comp->resolutions + l_tile_comp->minimum_num_resolutions - 1`.
When `minimum_num_resolutions == 0` this is `resolutions[-1]`, an out-of-bounds
read dereferenced at tcd.c:1426  the SEGV reported in #1631 (ASan, READ on an
unmapped address).

`minimum_num_resolutions` is computed in `opj_tcd_init_tile()` (tcd.c:880):

```c
if (l_tccp->numresolutions < l_cp->m_specific_param.m_dec.m_reduce) {
    l_tilec->minimum_num_resolutions = 1;
} else {
    l_tilec->minimum_num_resolutions =
        l_tccp->numresolutions - l_cp->m_specific_param.m_dec.m_reduce;   // 0 when equal
}
```

The intended invariant is `m_reduce < numresolutions` — COD parsing enforces it
in `opj_j2k_read_SPCod_SPCoc()` (j2k.c:10979) by rejecting
`m_reduce >= numresolutions`. But this guard uses `<`, so when
`m_reduce == numresolutions` it falls into the `else` branch and yields `0`.

`m_reduce` reaches that equal-to value through
`opj_j2k_set_decoded_resolution_factor()` (j2k.c:12556), which writes
`m_reduce = res_factor` **before** validating and then returns `OPJ_FALSE`:

```c
p_j2k->m_cp.m_specific_param.m_dec.m_reduce = res_factor;   /* mutated first */
...
    if (res_factor >= max_res) { ...; return OPJ_FALSE; }    /* rejected, but state already changed */
```

So a *rejected* `set_decoded_resolution_factor()` leaves the decoder in a
corrupted state, and a later `opj_read_tile_header()` trips the OOB.

### Reachability (honest scoping)
This is **not** reachable through `opj_decompress` or the in-tree OSS-Fuzz
harnesses (`opj_decompress_fuzzer_{J2K,JP2}`): those take the full-decode path,
where COD parsing rejects `m_reduce >= numresolutions` up front
(`opj_decompress -r 3` on a 3-resolution image exits cleanly:
*"The number of resolutions to remove (3) is greater or equal than the number of
resolutions of this component (3)"*). It is reachable from application code that
calls `opj_set_decoded_resolution_factor()` with an out-of-range factor and then
proceeds to `opj_read_tile_header()` without checking the `OPJ_FALSE` return.
Severity is correspondingly low  the standard decode path is unaffected — but a
public setter should not corrupt decoder state on its rejection path, and the
tcd.c guard is internally inconsistent with the COD-parse invariant.

### Fix
Two small, independent changes (either alone prevents the crash; together they
fix the root cause and harden the point of use):

1. **`opj_j2k_set_decoded_resolution_factor()` — root cause.** Validate every
   component before mutating any state, so a rejected factor is a no-op and
   leaves `m_reduce` and the per-component `factor` values unchanged. (This also
   closes a pre-existing partial-state issue: `comps[i].factor` was being set
   incrementally before a *later* component could fail validation.)

2. **`opj_tcd_init_tile()` — defense in depth.** Use `<=` so
   `minimum_num_resolutions` is clamped to `>= 1` whenever
   `m_reduce >= numresolutions`, matching the `>=` invariant the COD parser
   already enforces. This guarantees `opj_tcd_get_decoded_tile_size()` can never
   read `resolutions[-1]`, regardless of how `m_reduce` was set.

### Verification
Built with `-fsanitize=address` at `21b70b0`. A harness reproducing the #1631
sequence (`read_header` → `set_decoded_resolution_factor(rf)` →
`read_tile_header`) on the reported crash input:

| reduce factor              | before            | after                         |
|----------------------------|-------------------|-------------------------------|
| 1, 2 (valid)               | ok                | ok — identical tile sizes (13924 / 3600) |
| **3 (== numresolutions)**  | **SEGV @ tcd.c:1426** | **no crash, clean read**  |
| 4 (> numresolutions)       | ok (already clamped) | ok                         |

- **Each fix independently stops the SEGV.** With *only* the tcd.c clamp (setter
  left buggy, so `m_reduce` is still corrupted to 3), rf=3 no longer crashes 
  it clamps to `minimum_num_resolutions = 1`. With the setter fix, `m_reduce` is
  never corrupted in the first place.
- **No regression on the normal path:** `opj_decompress -r 0` / `-r 2` decode the
  file unchanged; `opj_decompress -r 3` still exits cleanly via the COD-parse
  check. Valid reduce factors (`m_reduce < numresolutions`) are byte-for-byte
  unaffected — the tcd.c change only diverges from current behavior in the
  `m_reduce == numresolutions` case, which valid decoding never reaches.

### DIFF
--- a/src/lib/openjp2/j2k.c
+++ b/src/lib/openjp2/j2k.c
@@ -12553,12 +12553,11 @@ OPJ_BOOL opj_j2k_set_decoded_resolution_factor(opj_j2k_t *p_j2k,
 {
     OPJ_UINT32 it_comp;
 
-    p_j2k->m_cp.m_specific_param.m_dec.m_reduce = res_factor;
-
     if (p_j2k->m_private_image) {
         if (p_j2k->m_private_image->comps) {
             if (p_j2k->m_specific_param.m_decoder.m_default_tcp) {
                 if (p_j2k->m_specific_param.m_decoder.m_default_tcp->tccps) {
+                    /* Validate against all components before mutating state. */
                     for (it_comp = 0 ; it_comp < p_j2k->m_private_image->numcomps; it_comp++) {
                         OPJ_UINT32 max_res =
                             p_j2k->m_specific_param.m_decoder.m_default_tcp->tccps[it_comp].numresolutions;
@@ -12567,6 +12566,9 @@ OPJ_BOOL opj_j2k_set_decoded_resolution_factor(opj_j2k_t *p_j2k,
                                           "Resolution factor is greater than the maximum resolution in the component.\n");
                             return OPJ_FALSE;
                         }
+                    }
+                    p_j2k->m_cp.m_specific_param.m_dec.m_reduce = res_factor;
+                    for (it_comp = 0 ; it_comp < p_j2k->m_private_image->numcomps; it_comp++) {
                         p_j2k->m_private_image->comps[it_comp].factor = res_factor;
                     }
                     return OPJ_TRUE;
@@ -12575,6 +12577,8 @@ OPJ_BOOL opj_j2k_set_decoded_resolution_factor(opj_j2k_t *p_j2k,
         }
     }
 
+    /* Image not available yet; validated later at COD parsing time. */
+    p_j2k->m_cp.m_specific_param.m_dec.m_reduce = res_factor;
     return OPJ_FALSE;
 }
 
--- a/src/lib/openjp2/tcd.c
+++ b/src/lib/openjp2/tcd.c
@@ -877,7 +877,8 @@ static INLINE OPJ_BOOL opj_tcd_init_tile(opj_tcd_t *p_tcd, OPJ_UINT32 p_tile_no,
         /*fprintf(stderr, "\tTile compo border = %d,%d,%d,%d\n", l_tilec->x0, l_tilec->y0,l_tilec->x1,l_tilec->y1);*/
 
         l_tilec->numresolutions = l_tccp->numresolutions;
-        if (l_tccp->numresolutions < l_cp->m_specific_param.m_dec.m_reduce) {
+        /* Use <= to prevent minimum_num_resolutions from underflowing to 0. */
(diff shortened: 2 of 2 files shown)


5asever40-a11y:
Hello,

Thank you for reproducing the issue and preparing the fix in PR #1640.

I saw that the PR describes the bug as an out-of-bounds read in
opj_tcd_get_decoded_tile_size() when the reduce factor equals
numresolutions,
and that the reachability is limited to an application/API path where
opj_set_decoded_resolution_factor() returns OPJ_FALSE but the caller
continues
to opj_read_tile_header().

I understand that this is low severity and not reachable through the normal
opj_decompress path. Still, since the issue affects a public API path,
causes
a reproducible ASan READ SEGV, and has a concrete fix, I wanted to ask:

Do the OpenJPEG maintainers plan to request or assign a CVE ID, or publish a
GitHub Security Advisory for this issue?

If the project does not plan to do so, would it be acceptable for us as the
reporters to request a CVE ID while clearly stating the limited reachability
and low severity?

We will make sure not to overstate the impact. The request would describe it
as a limited-reachability out-of-bounds read / denial-of-service condition,
not as code execution.

Thank you again for reviewing and fixing the report.

Best regards,

OH HAN GUEL
KANG DAEUN

2026년 5월 31일 (일) 오전 2:40, Sidhartha kumar ***@***.***>님이 작성:

PR #1644, opened 2026-06-10
opj_j2k: fix invalid tile part index error when decoding tiles with interleaved tile-parts

### LINKED ISSUE #1558: `Invalid tile part index` error on some JPEG 2000 files
## Expected behavior and actual behavior

Expected behavior: Calling `opj_get_decoded_tile` for each tile in an image should allow decoding of each tile in turn

Actual behavior: On some (not all) images, decoding fails with `Invalid tile part index for tile number 1. Got 7, expected 0`. This appears to be because of the reset of `m_current_tile_part_number` taking place [here](https://github.com/uclouvain/openjpeg/blob/master/src/lib/openjp2/j2k.c#L12211-L12218).

`opj_decompress` has no trouble at all with the image (either the full image or a single tile), and [jpylyzer](https://jpylyzer.openpreservation.org/) reports that the image is compliant with the JPEG 2000 spec.

The problem was first discovered trying to load certain jp2 files into libvips. The issue was reported and investigated in libvips/libvips#4205. There is much more detailed debugging output attached to that issue.

I managed to reproduce the issue outside libvips and create a minimal example.

Use this code and sample images: [tile_part_index_error.zip](https://github.com/user-attachments/files/17499121/tile_part_index_error.zip)

Compile and link `main.c` against libopenjp2, and then run the resulting binary with the name of the file to decode.

- `notworking.jp2` is the file that causes the error
- `working.jp2` does not cause the error. It is the exact same file decompressed and recompressed using:
    ```
    opj_decompress -i notworking.jp2 -o image.png
    opj_compress -i image.png  -o working.jp2 -t 1024,1024 -n 8
    ```

## Operating system

Both Linux (Fedora 36) and MacOS (Sequoia 15.0.1)

## openjpeg version

v2.5.2

mbklein:
Fixes #1558

### Problem
`opj_get_decoded_tile()` failed with "Invalid tile part index for tile number N. Got X, expected 0" on codestreams whose tile-parts are interleaved across tiles (all tiles' first tile-part, then all tiles' second tile-part, etc.) and that only declare the total tile-part count (`TNsot`) in each tile's last tile-part.

### Cause
When the tile is decoded, the prior tile decode stops as soon as its own last tile-part is consumed, which can be reached before the target tile's final tile-part (the one containing `TNsot`). The target tile's `nb_tps` is never set and its tile-part start positions are never recorded, so `opj_j2k_decode_one_tile()` takes the "index not built" fallback. That fallback seeks to `m_last_sot_read_pos`, which for an interleaved layout points past the tile's first tile-part, causing the decoder to encounter a non-zero `TPsot` first and abort.

This only occurs when multiple tiles are decoded from the same codec (e.g. `j2k_random_tile_access`). The first tile always succeeds because the index hasn't been built yet.

### Fix
- Seek to the start of the codestream (`main_head_end` + 2) instead, so the target tile's tile-parts are read in order. The faster index-based path used by ordered codestreams is unchanged.
- Add a self-contained non-regression test (`rta_interleaved_tile_parts`): it encodes a small multi-tile image, rewrites it into the interleaved layout, then decodes every tile and checks the samples against a full-image decode. The test fails before this fix and passes after it, with no external test data dependency.

### DIFF
--- a/src/lib/openjp2/j2k.c
+++ b/src/lib/openjp2/j2k.c
@@ -12185,10 +12185,17 @@ static OPJ_BOOL opj_j2k_decode_one_tile(opj_j2k_t *p_j2k,
     if (p_j2k->cstr_index->tile_index)
         if (p_j2k->cstr_index->tile_index->tp_index) {
             if (! p_j2k->cstr_index->tile_index[l_tile_no_to_dec].nb_tps) {
-                /* the index for this tile has not been built,
-                 *  so move to the last SOT read */
+                /* The index for this tile has not been built. This happens */
+                /* when the tile's tile-parts are interleaved with other */
+                /* tiles' tile-parts and the number of tile-parts (TNsot) was */
+                /* only declared in a tile-part not reached while decoding a */
+                /* previous tile. In that case we cannot rely on the recorded */
+                /* SOT positions, nor on m_last_sot_read_pos (which may point */
+                /* past the first tile-part of the requested tile), so seek */
+                /* back to the start of the codestream and read the tile-part */
+                /* headers in order. See issue #1558. */
                 if (!(opj_stream_read_seek(p_stream,
-                                           p_j2k->m_specific_param.m_decoder.m_last_sot_read_pos + 2, p_manager))) {
+                                           p_j2k->cstr_index->main_head_end + 2, p_manager))) {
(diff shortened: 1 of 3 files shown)


mbklein:
That one failing build action is due to a download hiccup, but it doesn't seem to be something I can retry myself.

PR #1646, opened 2026-06-11
ci: add permissions and pin action SHAs (4 workflows)

Alb3e3:
## Summary

The CI workflows are missing top-level \`permissions\` declarations and reference actions by mutable version tags, which are supply-chain risk vectors.

**Changes:**
- Add \`permissions: read-all\` at the workflow level for all workflows (least-privilege default)
- Pin all third-party action references to full commit SHAs

Mutable action references (\`@v4\`, \`@main\`, etc.) allow a compromised upstream action to run arbitrary code in your CI. Pinning to a commit SHA ensures only the audited version runs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    cooldown:
+      default-days: 7
+    groups:
+      all-actions:
+        patterns:
+          - "*"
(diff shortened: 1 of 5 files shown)


rouault:
I would want to see a dependabot config like https://github.com/OSGeo/gdal/blob/master/.github/dependabot.yml to update the SHAs

Alb3e3:
@rouault Done — I added a `.github/dependabot.yml` in `625b929` that enables the `github-actions` ecosystem on a monthly schedule (grouped, single PR for all actions), so Dependabot keeps the pinned SHAs updated automatically. Let me know if you'd prefer a different cadence or grouping.

Alb3e3:
@rouault Friendly ping — the dependabot config you asked for was added in `625b929` back on 2026-06-28 (`github-actions` ecosystem, monthly, grouped into a single PR so the pinned SHAs stay updated automatically). The branch is clean against master.

Happy to change the cadence or grouping if you would prefer something closer to the GDAL config.

PR #1657, opened 2026-07-12
Fix data loss / assert when encoding tiles with empty resolution levels

### LINKED ISSUE #1575: `opj_compress` sometimes loses data when encoding in lossless mode
Repro:

1. Download this input file:  [ref.zip](https://github.com/user-attachments/files/18630971/ref.zip) It contains ref.bmp.

It looks like so (in png form, for github's preview):

Image

2. Download osx-arm64 version from https://github.com/uclouvain/openjpeg/releases/tag/v2.5.3, fix up the rpath with `install_name_tool -add_rpath "@loader_path/../lib" ~/Downloads/openjpeg-v2.5.3-osx-arm64/bin/opj_compress`

3. Compress ref.bmp losslessly like so: `~/Downloads/openjpeg-v2.5.3-osx-arm64/bin/opj_compress -i ref.bmp -o openjpeg-lossless-rgba-u8-prog0-tile3x2-res5.jp2 -n 5 -t 39,53 -C "opj_compress -n 5 -t 39,53"`

4. Uncompress it again (or look at it in Preview.app): `~/Downloads/openjpeg-v2.5.3-osx-arm64/bin/opj_decompress -i openjpeg-lossless-rgba-u8-prog0-tile3x2-res5.jp2 -o out.png`

The output looks like so:

Image

Note the two white stripes at the far right.

This is probably related to the leftmost tiles being just 2 pixels wide. Since we have 4 decompositions, the LL band and the first decomposition after it are empty, which might confuse something.

nico:
When a tile is narrow but the number of decomposition levels is large, some of its lower resolution levels can become empty. Neither 5/3 not 9/7 DWT handled zero-length inputs correctly:

- opj_dwt_encode_and_deinterleave_h_one_row (5/3, lossless): The `if (even)` branch correctly checks `if (width > 1)`, but the else branch for the odd case special-cased `width == 1` and then just had an `else` branch on that. This caused width == 0 to be handled incorrectly. This is similar to the `if (even)` branch just above, and also similar to inverse opj_idwt53_h with its `if (len > 1)`. Fixes #1575.

- opj_dwt_encode_and_deinterleave_h_one_row_real (9/7, lossy): only returned early for width == 1, so for width == 0 it called opj_dwt_encode_1_real() with dn == sn == 0. This tripped the `assert(dn + sn > 1)` therein in debug builds. Return early for `width <= 1`, to handle empty tiles too.

Before this commit, reproduced with a 119x53 grayscale image: width 119 with `-t 117` leaves a 2-pixel-wide right edge tile, and `-n 5` makes that tile's three lowest resolution levels empty:

    # lossless: cols 117-118 corrupted
    opj_compress -i in.pgm -o out.jp2 -n 5 -t 117,53

    # lossy: assert abort in debug builds
    opj_compress -i in.pgm -o out.jp2 -n 5 -t 117,53 -I

### DIFF
--- a/src/lib/openjp2/dwt.c
+++ b/src/lib/openjp2/dwt.c
@@ -1361,7 +1361,7 @@ void opj_dwt_encode_and_deinterleave_h_one_row(void* rowIn,
     } else {
         if (width == 1) {
             row[0] *= 2;
-        } else {
+        } else if (width > 1) {
             OPJ_INT32 i;
             tmp[sn + 0] = row[0] - row[1];
             for (i = 1; i < sn; i++) {
@@ -1393,7 +1393,7 @@ void opj_dwt_encode_and_deinterleave_h_one_row_real(void* rowIn,
     OPJ_FLOAT32* OPJ_RESTRICT tmp = (OPJ_FLOAT32*)tmpIn;
     const OPJ_INT32 sn = (OPJ_INT32)((width + (even ? 1 : 0)) >> 1);
     const OPJ_INT32 dn = (OPJ_INT32)(width - (OPJ_UINT32)sn);
-    if (width == 1) {
+    if (width <= 1) {
         return;
     }
     memcpy(tmp, row, width * sizeof(OPJ_FLOAT32));


nico:
I have a roundtrip test for this too, but I'll put it in a separate PR in case the fix is easier to merge than the test for some reason.

nico:
@rouault Mind merging this?

fallenmi (approved):
APPROVE

I verified this on exact head `5267c2a22fdb0f1334683fd69a9df118ae7ceace` and on the current clean merge. A Debug-build matrix covered six right-edge tile splits in both the horizontal and rotated vertical orientations, with lossless roundtrip equality and lossy encode completion for each case. The exact base passes only 13/24 checks (the motivating horizontal cases either corrupt the lossless output or hit `opj_dwt_encode_1_real`'s `dn + sn > 1` assertion); the head and current merge pass 24/24.

I also checked RGB inputs with four encoder threads in both orientations and modes (8/8 across head and merge), and the focused tile/area/random-access/unit set passes 29/29 locally runnable tests on the head and 30/30 on the current merge. The one additional head-side `testjp2` target requires the separate `openjpeg-data` checkout and was not runnable because `OPJ_DATA_ROOT` was unset. The two-line guards match the existing even-branch and inverse-transform handling, and the rotated zero-height controls remain unchanged and correct.

This review was prepared with OpenAI Codex assistance under the account owner's authorization. The exact diff, source paths, merge result, and test evidence above were inspected and verified in the account owner's Codex session before submission.

PR #1661, opened 2026-08-10
t2/tcd: fix precinct OOB access and NULL tag-tree dereference

Alb3e3:
## Summary

Two memory-safety bugs in the JPEG 2000 packet-header decoder and tile initialiser.

---

### Bug A — OOB precinct array access in `opj_t2_read_packet_header` (HIGH)

**File:** `src/lib/openjp2/t2.c`

Two sites computed `&l_band->precincts[p_pi->precno]` before (or without) a bounds check:

1. **layno == 0 early path** (~original line 1087): the pointer was computed on the line immediately before the existing bounds gate, so the guard came too late to prevent the C11 §6.5.6p8 undefined-behaviour pointer arithmetic.

2. **Main band decode loop** (~original line 1200): no bounds check existed at all. A crafted COC marker that widens the precinct grid after the initial tile allocation could push `precno` past the allocated array, enabling a heap out-of-bounds read/write.

**Fix:** in both sites the pointer variable is now declared before the guard and assigned only after the bounds check passes. The error message also reports the offending index and the array limit.

---

### Bug B — NULL dereference from `opj_tgt_create` / `opj_tgt_init` in `opj_tcd_init_tile` (MEDIUM)

**File:** `src/lib/openjp2/tcd.c`

`opj_tgt_create` and `opj_tgt_init` both document that they can return NULL on allocation failure (large precincts may trigger allocations of hundreds of MB). The return values for `incltree` and `imsbtree` were not checked; NULL stored here would cause an immediate NULL dereference inside `opj_tgt_reset` on the next tile decode cycle.

**Fix:** after the create/init pair for each tree, verify both pointers are non-NULL and return `OPJ_FALSE` with an `EVT_ERROR` message if either allocation failed.

---

## Testing

- Library and tools build cleanly (no new warnings or errors).
- All 26 CTest tests pass (`tte*`, `ttd*`, `rta*`, `tda*`).


Alb3e3:
The AppVeyor failure on this branch was my bug, not flakiness. Fixed in `d6ce0604`.

`opj_tgt_create()` returns NULL when the tag tree would have zero nodes, which is exactly the case for a precinct with no code-blocks (`cw` or `ch` is 0). That is legal — the trees are simply never used for such a precinct, and `opj_tgt_reset()` is already NULL-safe. My unconditional NULL check turned that normal state into a decode error:

```c
-                    if (!l_current_precinct->incltree ||
-                            !l_current_precinct->imsbtree) {
+                    if (l_nb_code_blocks != 0 &&
+                            (!l_current_precinct->incltree ||
+                             !l_current_precinct->imsbtree)) {
```

Since `l_nb_code_blocks == cw * ch`, gating on it distinguishes a genuine allocation failure from a legitimately empty precinct.

Confirmed with an instrumented build — the empty case is real and common:

```
DBG precno=0 cw=1 ch=1 nbcblk=1 incl=0x...780 imsb=0x...870
DBG precno=0 cw=0 ch=1 nbcblk=0 incl=(nil)  imsb=(nil)     <-- legal, was rejected
```

Measured against `openjpeg-data/input/conformance`, decoding all 44 streams with each build of `opj_decompress`:

| build | decode failures | tag-tree errors |
| --- | --- | --- |
| as previously pushed (`b733cd9a`) | **9 / 44** | 9 |
| with this fix (`d6ce0604`) | **0 / 44** | 0 |

The nine were `p0_12`, `p0_13`, `p1_05`, `p1_06`, `p1_07`, `b3_mono`, `d2_colr`, `e1_colr`, `e2_colr` — which matches the AppVeyor run, where the same set expands to ~45 test cases (97% passed, 47 failed vs a 2-failure baseline on green builds).

The `t2.c` precinct bounds checks in the first commit are unchanged and were not implicated.

